PR #1, opened 2020-05-26
Search for double-fuzzy files only when needed

benrubson:
Hi,

This PR improves performance when using double-fuzzy (`-yy`).
It browses the `--link-dest` directories only if needed, only when it has to look for a basis-file.
This then saves time / resources / improves performance when only simple hard-links are needed.

Many thanks 👍

_Initial report [there](https://bugzilla.samba.org/show_bug.cgi?id=12498), pushed here by convenience._

### DIFF
--- a/generator.c
+++ b/generator.c
@@ -1315,21 +1315,6 @@ static void recv_generator(char *fname, struct file_struct *file, int ndx,
 		}
 		parent_dirname = dn;
 
-		if (need_fuzzy_dirlist && S_ISREG(file->mode)) {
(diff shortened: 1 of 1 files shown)


WayneD:
Merged via patch (since I like the straight-forward history that results that way).

benrubson:
I wonder whether or not one of the 3 merge options would result in a straight-forward history :

image

From the repository options :

image

For sure, thank you for this merge :)

kuntawan (approved):

PR #2, opened 2020-05-26
Correctly send last error to sender

benrubson:
Hi,

Here is a PR which sends last error messages to sender for a file-transfer which failed.
It helps especially with protocol < 31.
This can be useful for example to know that a FS is full.

Many thanks 👍

```
### Without :

/repository/mybigfile.iso
        539.26M   6%   11.21MB/s    0:11:01  
rsync: [sender] write error: Broken pipe (32)
rsync error: error in file IO (code 11) at io.c(831) [sender=3.1.2]

### With (of course the interesting message here is the first one) :

/repository/mybigfile.iso
        539.26M   6%   11.21MB/s    0:11:01  
rsync: write failed on "/repository/mybigfile.iso" (in bkp): Disc quota exceeded (69)
rsync error: error in file IO (code 11) at receiver.c(400) [receiver=3.1.2]
rsync error: error in file IO (code 11) at io.c(1633) [generator=3.1.2]
rsync: [sender] write error: Broken pipe (32)
rsync error: error in file IO (code 11) at io.c(831) [sender=3.1.2]
```

_Initial report [there](https://bugzilla.samba.org/show_bug.cgi?id=12522), pushed here by convenience._

### DIFF
--- a/cleanup.c
+++ b/cleanup.c
@@ -26,6 +26,7 @@ extern int dry_run;
 extern int am_server;
 extern int am_daemon;
 extern int am_receiver;
+extern int am_sender;
 extern int io_error;
 extern int keep_partial;
 extern int got_xfer_error;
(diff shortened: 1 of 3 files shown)


benrubson:
Well, we could simply ignore this PR, especially because everything works fine with protocol 31, which is now almost [7 years old](https://github.com/WayneD/rsync/blob/master/OLDNEWS.md#news-for-rsync-310-28-sep-2013).
But this could be useful for PR #3, so I keep it opened for now.

benrubson:
Many thanks Wayne :+1:

PR #4, opened 2020-05-26
Sync once files received

benrubson:
Hi,

Here is a PR which simply `sync()` once files received, for data consistency.

Many thanks 👍

_Initial report [there](https://bugzilla.samba.org/show_bug.cgi?id=12819), pushed here by convenience._

### DIFF
--- a/main.c
+++ b/main.c
@@ -1006,6 +1006,7 @@ static int do_recv(int f_in, int f_out, char *local_name)
 		io_start_multiplex_out(f_out);
 
 		recv_files(f_in, f_out, local_name);
+		sync(); /* data consistency, force completion of pending disk writes */
 		io_flush(FULL_FLUSH);
 		handle_stats(f_in);
 


WayneD:
I don't like the idea of forcing all pending data from anyone's disk activity to disk just because someone ran an rsync. For a remote daemon someone could put a sync into the "post-xfer exec" script. For a receiver via ssh, I'd suggest creating a shell script and running it via --rsync-path=rsync-and-sync when you want the syncing to be done.

#!/bin/bash
# maybe check if we're the --sender and avoid the sync?
# maybe save rsync's ret value & always do the sync & then return the ret value?
rsync "${@}" && sync

benrubson:
OK Wayne, thank you for your feedback.
I liked the idea that, as a backup program, _rsync_ replied with "_OK : backup successfully finished, and your data is safely on disk_".
I'll use a workaround then :)
Thank you !

PR #6, opened 2020-05-31
MD5-related optimizations

Chainfire:
- This is a 3-parter. The first commit moves the OpenSSL related defines from from checksum.c to mdigest.h, because I found myself copy/pasting them more than once otherwise.

- The second commit enables parallel computation of MD5 hashes in the block matching get_checksum2() phase. As each blocks' hash is independent, we can process up to 4 blocks simultaneously with SSE2, and 8 blocks with AVX2, leading to a real-world 2x to 6x performance gain / CPU usage reduction (even over OpenSSL-optimized MD5). 

However, to make this happen without significant changes to the rest of rsync's codebase, a block prefetcher had to be created. (This whole commit requires --enable-simd as my previous contributions). Full compatibility is maintained with non-SIMD counterparts.

The same mechanism could be used for multithreading checksums as well, but that is beyond the scope of this patch.

- The third commit provides the MD5P8 whole-file checksum. This is an MD5-based checksum but cuts the input stream into 8 independent streams (64-byte interleave), of which the final states are brought together to create the final MD5 checksum. This has the same strengths (and weaknesses) of the normal MD5 checksum, but allows parallel processing, with again a 2x to 6x performance gain. 

Both optimized parallel processing is available (--enable-simd) as reference C. MD5P8 is slightly slower on <10kB files due to the additional overhead, but similar to MD5 on larger files without SIMD, and much faster on larger files with SIMD.

Note that get_checksum2() keeps using normal MD5 even if whole-file checksum is MD5P8, because that is parallelized with SIMD anyway if available, and using MD5P8 would just add overhead and quite probably be slower.

Further note that the CSUM_MD5 and CSUM_MD5P8 defines now appear in both checksum.c and simd-checksum-x86_64.cpp, they need to be kept in sync, perhaps moved to a header?

----

Motivation: though xxhash is now available for rsync, it is not included into the code itself but an external dependency, and by my last evaluation, many distros do not yet come with xxhash included, and thus the distro-included rsync package will be built without xxhash support. That being said, I can imagine that this PR may not be merged due to it not being part of the direction rsync is moving in, I myself need to be using it due to having an uncommon build target, and the code might as well be available to everyone.

The parallel computation of MD5 hashes in get_checksum2() will benefit connections to both recent builds without xxhash as well as older builds of rsync if the block-matching phase applies. If it doesn't lead to a reduction in transfer time due to connection or disk speed limitations, then it will at least massively reduce CPU usage on the supporting client.

The use-case for MD5P8 is more limited, as its usefulness requires both ends to be running an supporting rsync build, but one end not supporting xxhash. If both ends do support xxhash, that should always be the preferred checksum (while MD5P8 can reach gigabytes per second, xxhash is still twice as fast). I only created it as it was a small effort now that parallel MD5 computation was available anyway, and it doesn't have any external dependencies.

----

I've done some benchmarks for transferring 1GB files between a fast and a slow CPU on 1GbE LAN, compared to normal MD5 usage (all tests already including my previous block size patches and get_checksum1() optimizations):

get_checksum2() MD5 parallelization with MD5 whole-file checksum, both files existing on both ends:
- 33% transfer time reduction
- 52% CPU usage reduction

get_checksum2() MD5 parallelization and MD5P8 whole-file checksum, both files existing on both ends:
- 54% transfer time reduction
- 84% CPU usage reduction

xxhash for both get_checksum2() and whole-file checksum, both files existing on both ends:
- 54% transfer time reduction
- 90% CPU usage reduction

MD5P8, new file:
- 33% transfer time reduction
- 86% CPU usage reduction

xxhash, new file:
- 33% transfer time reduction
- 92% CPU usage reduction

MD5P8, local checksum:
- 83% CPU usage reduction

xxhash, local checksum:
- 94% CPU usage reduction

Obviously these are highly specific to my setup and YMMV. However, my daily syncing of TB's of data is now twice as fast, with average CPU usage down to less than a quarter. xxhash doesn't run ahead much in this case because CPU power while checksumming is no longer the bottleneck after these patches. With even faster network and disks (10GbE + NVMe) xxhash might be twice as fast.

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -31,7 +31,7 @@ VERSION=@RSYNC_VERSION@
 .SUFFIXES:
 .SUFFIXES: .c .o
 
-SIMD_x86_64=simd-checksum-x86_64.o lib/md5-asm-x86_64.o
+SIMD_x86_64=simd-checksum-x86_64.o simd-md5-parallel-x86_64.o lib/md5-asm-x86_64.o
 
 GENFILES=configure.sh aclocal.m4 config.h.in proto.h proto.h-tstamp rsync.1 rsync-ssl.1 rsyncd.conf.5
 HEADERS=byteorder.h config.h errcode.h proto.h rsync.h ifuncs.h itypes.h inums.h \
@@ -120,6 +120,9 @@ rounding.h: rounding.c rsync.h proto.h
 	@rm -f rounding.out
 
 simd-checksum-x86_64.o: simd-checksum-x86_64.cpp
+	$(CXX) $(CXXFLAGS) $(CPPFLAGS) -c -o $@ $<
+
+simd-md5-parallel-x86_64.o: simd-md5-parallel-x86_64.cpp
 	$(CXX) $(CXXFLAGS) $(CPPFLAGS) -c -o $@ $<
 
 lib/md5-asm-x86_64.o: lib/md5-asm-x86_64.s
--- a/Makefile.in
+++ b/Makefile.in
@@ -36,7 +36,7 @@ SIMD_x86_64=simd-checksum-x86_64.o simd-md5-parallel-x86_64.o lib/md5-asm-x86_64
 GENFILES=configure.sh aclocal.m4 config.h.in proto.h proto.h-tstamp rsync.1 rsync-ssl.1 rsyncd.conf.5
 HEADERS=byteorder.h config.h errcode.h proto.h rsync.h ifuncs.h itypes.h inums.h \
 	lib/pool_alloc.h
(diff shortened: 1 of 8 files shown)


WayneD:
Thanks!  I've put the changes into a file named "md5p8.diff" in the rsync-patches repo for now. I incorporated some of the changes that put more info into lib/mdigest.h, and I tweaked a few things for style and to fix a compiler warning. Here's the resulting patch:

https://git.samba.org/?p=rsync-patches.git;a=blob_plain;f=md5p8.diff;hb=ac98f867ff5e7e53a0157b967c7b216c86b0b0a6

WayneD:
I'm going to leave it as a maintained patch for now and consider merging it later.

Chainfire:
I'll update this with the new build tests and applying to latest master

PR #9, opened 2020-06-07
zstd: Use `0' as default compression level

sebastianas:
The default compression level for zstd is defined by ZSTD_CLEVEL_DEFAULT
and is currently `3'.

Use the special value `0' (which is controlled by ZSTD_CLEVEL_DEFAULT)
instead of `3'.

Signed-off-by: Sebastian Andrzej Siewior <sebastian@breakpoint.cc>


WayneD:
No, 0 turns into 3, so 3 is clearer to the user.

PR #12, opened 2020-06-15
fix typo in NEWS

eworm-de:


WayneD:
Thanks for the report. I fixed the spelling when I was checking in a few other minor tweaks.

PR #18, opened 2020-06-17
enhance and add systemd units

eworm-de:
Note this does limit rsync via systemd - it's a trade off for security.
With these units system (with some exceptions) is read-only, users' homes and devices are hidden. Comments give a hint how to undo this.


eworm-de:
Let me know if you think this is too strict...

Issue #13, opened 2020-06-16
build fails with SIMD checksum optimizations


eworm-de:
Building 3.2.0pre2 in current Arch Linux environment I get:

```
g++ -march=x86-64 -mtune=generic -O2 -pipe -fno-plt -D_FORTIFY_SOURCE=2 -c -o simd-checksum-x86_64.o simd-checksum-x86_64.cpp
{standard input}: Assembler messages:
{standard input}:280: Error: symbol `_ZL21get_checksum1_avx2_64PaiiPjS0_' is already defined
{standard input}:301: Error: symbol `_ZL21get_checksum1_sse2_32PaiiPjS0_' is already defined
make: *** [Makefile:128: simd-checksum-x86_64.o] Error 1
```

benrubson:
Some issues on Debian9 with SIMD :
```
g++ -g -O2  -c -o simd-checksum-x86_64.o simd-checksum-x86_64.cpp
simd-checksum-x86_64.cpp: In function ‘int32_t get_checksum1_sse2_32(signed char*, int32_t, int32_t, uint32_t*, uint32_t*)’:
simd-checksum-x86_64.cpp:161:40: error: ‘__m128i_u’ was not declared in this scope
         __m128i ss1 = _mm_loadu_si128((__m128i_u*)x);
                                        ^~~~~~~~~
simd-checksum-x86_64.cpp:161:50: error: expected primary-expression before ‘)’ token
         __m128i ss1 = _mm_loadu_si128((__m128i_u*)x);
                                                  ^
simd-checksum-x86_64.cpp:163:50: error: expected primary-expression before ‘)’ token
         __m128i ss2 = _mm_loadu_si128((__m128i_u*)x);
                                                  ^
simd-checksum-x86_64.cpp:166:53: error: expected primary-expression before ‘)’ token
         __m128i mul_t1 = _mm_loadu_si128((__m128i_u*)mul_t1_buf);
                                                     ^
simd-checksum-x86_64.cpp:178:52: error: expected primary-expression before ‘)’ token
                 in8_1 = _mm_loadu_si128((__m128i_u*)&buf[i]);
                                                    ^
simd-checksum-x86_64.cpp:179:52: error: expected primary-expression before ‘)’ token
                 in8_2 = _mm_loadu_si128((__m128i_u*)&buf[i + 16]);
                                                    ^
simd-checksum-x86_64.cpp:181:51: error: expected primary-expression before ‘)’ token
[...]

PR #20, opened 2020-06-18
x86-64 SIMD build fixes

Chainfire:
configure.ac was modified to detect g++ 5+ and clang++ 7+. Additionally
some script malfunctions on FreeBSD were corrected.

The get_checksum1() code has been modified to fix clang and g++ 10
compilation.

This version of the code and configure.ac has been tested on:

Ubuntu 16 - gcc 7.3.0, clang 6.0.0
Debian 10 - gcc 5.4.0, 6.4.0, 7.2.0, 8.4.0, 9.2.1, 10.0.1, clang 5.0.2,
6.0.1, 7.0.1, 8.0.0, 9.0.0, 10.0.0
ArchLinux 20200605 - gcc 10.1.0, clang 10.0.0
FreeBSD 12.1 - gcc 9.3.0, clang 8.0.1

It is unknown if it will work on gcc 5.0-5.3, but the script currently
allows it.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -200,48 +200,45 @@ AC_ARG_ENABLE(simd,
     AS_HELP_STRING([--enable-simd],[enable SIMD optimizations (requires c++)]))
 
 if test x"$enable_simd" = x"yes"; then
-    # For x86-64 SIMD, g++ 5+ or clang++ 7+ is required
+    # For x86-64 SIMD, g++ >=5 or clang++ >=7 is required
     if test x"$build_cpu" = x"x86_64"; then
-        if test x"$CXX" != x""; then
-            CXX_OK=
-            CXX_VERSION=$($CXX --version | head -n 1)
-            if `echo $CXX_VERSION | grep "^g++" >/dev/null`; then
-                CXX_BRAND="g++"
-            elif `echo $CXX_VERSION | grep "clang" >/dev/null`; then
-                CXX_BRAND="clang++"
-            fi
+	if test x"$CXX" != x""; then
+	    CXX_OK=
+	    CXX_VERSION=`$CXX --version | head -n 1`
+	    case "$CXX_VERSION" in
+	    g++*)
+		CXX_VERSION=`$CXX -dumpversion | sed 's/\..*//g'`
+		if test "$CXX_VERSION" -ge "5"; then
+		    CXX_OK=yes
+		fi
+		;;
+	    clang*)
+		# $CXX -dumpversion would have been ideal, but is broken on older clang
+		CXX_VERSION=`echo "$CXX_VERSION" | sed 's/.*version //g' | sed 's/\..*//g'`
+		if test "$CXX_VERSION" -ge "7"; then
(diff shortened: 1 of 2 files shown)


Chainfire:
See #13

hhoffstaette:
Confirmed that SIMD now builds with gcc-10 / clang-10 on Gentoo without having to futz with -march. Great work, thank you!

hhoffstaette:
Also thanks for the built-in benchmark/sanity check, it clearly shows that everything works correctly and with multiversion selection. :smiley:

eworm-de:
I can confirm this builds in a clean Arch Linux build environment with default flags. Thanks a lot!

Chainfire:
> Can confirm that SIMD now builds with gcc-10 / clang-10 using various optimizations on Gentoo amd64 and without having to futz with -march flags, despite building on an older non-AVX2 CPU. Great work, thank you!

Thanks for pointing out the problems. Especially GCC 10 is an interesting case, as it has problems not present on GCC 5-9. There's a regression in GCC 10's multiversion function handling, I should probably submit that to the GCC maintainers.

> Also thanks for the built-in benchmark/sanity check, it clearly shows that everything works correctly and with multiversion selection. 😃

No worries. I had the code externally, thought I might as well put it in. Depending on your configure flags it might not work but I guess anyone who really wants to test can figure it out.

WayneD:
Keep in mind that the shell that is running configure cannot be assumed to be bash, so things like `$(...)` instead of \`...\` are not allowed. I'm editing those out.

hhoffstaette:
> Keep in mind that the shell that is running configure cannot be assumed to be bash

This made me curious,so I checked and `$(..)` **is** POSIX, see the [bash FAQ](http://mywiki.wooledge.org/BashFAQ/082) on this.

Chainfire:
> Keep in mind that the shell that is running configure cannot be assumed to be bash, so things like `$(...)` instead of `...` are not allowed. I'm editing those out.

I did not know this; but the FreeBSD I tested on doesn't even *have* bash and it worked fine.

Either way, seems reasonable :)

Chainfire:
Nitpicking maybe, but just in case clang ever changes their description to clang++ in the output, shouldn't the clang case come first? This is why my submit checked for ^g++ rather than just g++.

WayneD:
The match is anchored at the start of the string.

WayneD:
Thanks for the fixes!  Seems like it's a good deal more portable now.

Chainfire:
I see. In that case it breaks FreeBSD clang compilation, which uses this version string:

`FreeBSD clang version 8.0.1 (tags/RELEASE_801/final 366581) (based on LLVM 8.0.1)`

Hence my grep for clang *not* having ^. I did just test this, and it does break.

Built successfully on all the other mentioned distros/gcc/clang combinations mentioned in the commit message (if they satisfy version requirements).

WayneD:
Oops, I missed that it was unanchored. Fixed.

Issue #21, opened 2020-06-19
Deprecated xattr header warnings 


hhoffstaette:
A few years ago xattr changed their headers to be provided by glibc (on Linux/glibc at least).
Right now the old/cross-platform way of checking for various xattr headers in `lib/sysxattrs.h`
results in the following warning:
```
clang -I. -I. -pipe -march=native -O3 -DHAVE_CONFIG_H -Wall -W  -c lib/sysxattrs.c -o lib/sysxattrs.o
In file included from lib/sysxattrs.c:23:
In file included from lib/sysxattrs.h:4:
/usr/include/attr/xattr.h:5:2: warning: "Please change your <attr/xattr.h> includes to <sys/xattr.h>" [-W#warnings]
#warning "Please change your <attr/xattr.h> includes to <sys/xattr.h>"
 ^
1 warning generated.
```

WayneD:
Merged the fix.

PR #22, opened 2020-06-19
Prevent unnecessary xattr warning by reordering header inclusion.

hhoffstaette:
xattr headers have been provided by glibc (at least on Linux/glibc) for many years now.
Reorder the inclusion of xattr headers to attempt compatibility/legacy after the common case.
This prevents the warning without changing compatibility to old/non-glibc systems.


hhoffstaette:
Fixes #21

PR #24, opened 2020-06-20
configure.ac: Improve check for g++

Polynomial-C:
Gentoo Linux has `x86_64-pc-linux-gnu-g++` (or similar string for other arches)
in CXX variable when gcc is being used and thus the check always fails.

Signed-off-by: Lars Wendler <polynomial-c@gentoo.org>


WayneD:
Since the check got changed to match the --version output I had assumed that g++ always started its --version string with "g++", but it apparently starts it with the basename from argv[0]. I changed the case match to go back to matching $CXX and it now checks `g++|*-g++)` (which I like better than `*g++)`.  I also tweaked a few other things in the matching logic while I was at it.  Sorry that this didn't make it into 3.2.0.

PR #28, opened 2020-06-20
Pass --noexecstack to assembler.

hhoffstaette:
This prevents Linux from rightfully complaining about an executable stack segment, which is widely considered a security hazard.


hhoffstaette:
Fixes #28

WayneD:
Change included manually.

PR #31, opened 2020-06-20
lib/md5-asm-x86_64.S: fix build on other macOS compilers

alebcay:
Follow-up from #25. Looks like the issue isn't just with Apple's clang; it affects other compilers on macOS as well. Adjust the `ifndef` to just check if OS is macOS rather than check if compiler is Apple clang.

### DIFF
--- a/lib/md5-asm-x86_64.S
+++ b/lib/md5-asm-x86_64.S
@@ -32,7 +32,7 @@
 .text
 .align 16
 
-#ifndef __apple_build_version__
+#ifndef __APPLE__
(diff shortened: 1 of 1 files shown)


WayneD:
I don't see why we even need those directives at all anymore with the ifdefs that get rid of the code when it's unneeded.

alebcay:
I'm not well-versed in assembly to know for sure, but I'm fairly certain that the directives are expected in the Linux (ELF) x86_64 format. If it compiles and works just fine without it though, I guess why not. 😄

PR #36, opened 2020-06-22
Add manpage explanation for --compres-level values

samueloph:
Hello Wayne,

I'm sending this patch we currently apply in the Debian packaging of rsync, I believe it would be nice to have it upstreamed.

Thanks,

### DIFF
--- a/rsync.1.md
+++ b/rsync.1.md
@@ -2314,8 +2314,9 @@ your home directory (remove the '=' for that).
 0.  `--compress-level=NUM`
 
     Explicitly set the compression level to use (see `--compress`) instead of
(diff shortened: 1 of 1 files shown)


WayneD:
Sadly, this detail is no longer true since it only applies to zlib compression.

samueloph:
@WayneD thanks, that's good to know, I will remove this one from the Debian package.

WayneD:
I've fleshed out that section of the man page with up-to-date details for 3.2.1.

samueloph:
Awesome! Looking forward for 3.2.1

PR #54, opened 2020-07-09
Add FreeBSD CI

benrubson:
Hi,

This is a very first version of a FreeBSD CI.
It's based on Cirrus CI : https://cirrus-ci.org
Used in other projects, it works perfectly.

Once we have this build properly working, goal is then to include tests, as in the Linux CI.

You should then have to enable Cirrus CI on this repository Wayne, if you're OK to go further with this.

Many thanks 👍

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -1,0 +1,10 @@
+task:
+  name: FreeBSD
+  freebsd_instance:
+    image: freebsd-12-1-release-amd64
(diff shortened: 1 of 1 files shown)


benrubson:
Many thanks for this merge @WayneD !
I was not sure about the default packages in the image, I had planned to correct accordingly, but glad to see you managed to get a successful build :+1:

PR #55, opened 2020-07-10
Clarify units in manual

paulmenzel:


WayneD:
Thanks for pointing that out.  I've fixed it in my own commit.

paulmenzel:
As the issue wasn’t mentioned in the commit message, this is fixed by commit https://github.com/WayneD/rsync/commit/7d63f8b249863fed83f82d97324d67ed5b141889.

PR #59, opened 2020-07-11
Enable FreeBSD CI tests

benrubson:
Hi,

This follows #54, enabling FreeBSD CI tests.

Thank you 👍


WayneD:
I tweaked it with a few more script steps instead.

PR #61, opened 2020-07-12
Improve FreeBSD tests

benrubson:
Hi,

Let's try to improve FreeBSD tests :
- mute the `acl_get_file() failed: Operation not supported` errors in non-acl tests ;
- use the correct attr commands in tests using them ;
- create a ZFS pool and mount it on `testtmp` dir, so that tests will run in this dedicated FS, which then makes `chmod-temp-dir` to run, as well as `dir-sgid` (on UFS, sticky bit is not inherited, whereas it is on ZFS).

Thx 👍

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -3,7 +3,8 @@ task:
   freebsd_instance:
     image: freebsd-12-1-release-amd64
   prep_script:
-    - mount -t tmpfs -o size=128m tmpfs /var/tmp
+    - dd if=/dev/zero of=/tmp/zpool bs=1M count=1024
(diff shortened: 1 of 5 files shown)


benrubson:
So there's the SGID test remaining, to clarify perhaps...
And the ACLs ones, would perhaps need to have a FS mounted with `-o acls`.

benrubson:
So there are the ACLs tests remaining.
We may have to tune the commands used :
```
# setfacl -m u:0:7 test
setfacl: test: branding mismatch; existing ACL is NFSv4, entry to be merged is POSIX.1e
```

WayneD:
Very nice improvements there!

PR #64, opened 2020-07-13
Enable FreeBSD python

benrubson:
Hi,

This enables python on FreeBSD.

Thx 👍

PR #65, opened 2020-07-14
Enable FreeBSD python

benrubson:
Hi,

This is another test for FreeBSD python.

Thx 👍


benrubson:
OK, here's another solution @WayneD.
Difference with #64 is that we do not install `git`, which takes time.
I also re-enabled `rsync-ssl` test, as `bash` is already built-in.

WayneD:
It's good to know what it will take to get docs going on FreeBSD (which I was curious about), but I'm going to leave it disabled. It's a good test of the `--disable-md2man` logic.

PR #67, opened 2020-07-14
Add Cygwin CI

benrubson:
Hi,

This PR adds a Cygwin CI.
`xxhash` is not available for now, but we should be able to add it anytime soon :
https://cygwin.com/pipermail/cygwin-apps/2020-July/040310.html

It uses [Travis](https://travis-ci.org), I successfully built Cygwin CI for another project using this, so I went for the same solution.

Thx 👍

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,15 @@
+matrix:
+  include:
+    - language: bash
+      os: windows
+      script:
+      - choco uninstall -y mingw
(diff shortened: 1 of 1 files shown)


WayneD:
In order to avoid adding yet another CI service, I put cygwin into Cirrus for now.  I'm thinking that using https://github.com/marketplace/actions/chocolatey-action might be even better, though.

PR #68, opened 2020-07-15
Force git line endings

benrubson:
Hi,

If we want to build a Windows/Cygwin CI, we'll need to tell git not to transform line endings.
I simply forgot it in #67, while I clearly needed it in other projects' Cygwin CI.

Sounds like you also faced the `\r` issue in some of your CI tests @WayneD.
This will then help.

Ben

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+* text=auto eol=lf


113ll3 (commented):
.gitattributes

PR #69, opened 2020-07-15
Let's try cygwin again.

WayneD:


.github/workflows/build.yml
```diff
@@ -52,3 +52,27 @@ jobs:
...
+    - name: make
+      run: c:/tools/cygwin/bin/bash -c 'make'
+    - name: install
+      run: bash -c 'make install'
+    - name: info
+      run: bash -c '/usr/local/bin/rsync --version'
+    - name: check
+      run: bash -c 'make check'
+    - name: ssl file list
+      run: bash -c '/usr/local/bin/rsync-ssl --no-motd download.samba.org::rsyncftp/ || true'
```
benrubson, line 78: Be perhaps sure to call `bash` with full path ?
As there's another `bash` on the system, and the Cygwin one is certainly the last one in `$PATH`.

benrubson:
Strangely the test which was failing in your first commit was properly running on my development environment.
Perhaps because I run tests as administrator ?
Perhaps because you don't use `bash --login` ? (thus skipping the Cygwin login procedure)
Seems strange :)

WayneD:
Yeah, I had not seen that test fail before either.  The test failure was a combination of the checkout files being owned by a different user than the CI test user combined with the fact that the CI user could preserve the owner with `cp -p` combined with the fact that rsync doesn't attempt to save the ownership as a non-root UID (since --super wasn't specified).

PR #73, opened 2020-07-22
Use RSYNC_MAX_SKIPPED on FreeBSD CI

benrubson:
Hi,

Let's also use `RSYNC_MAX_SKIPPED` on FreeBSD CI.

Thx 👍


WayneD:
Sure, that's a good idea.  I tossed this change into a few other build changes I made.

PR #75, opened 2020-07-25
NEWS.md typo

benrubson:


WayneD:
Thanks very much for the error correction (I really appreciate that I'm not the only one looking things over).  However, after a good night's sleep I decided to simplify the options a bit, so I've changed all 3 stderr-related options to `--stderr=MODE`.

PR #99, opened 2020-09-16
Add check for truncation in append mode

### LINKED ISSUE #90: Rsync segfaults with --append option when source file gets highly truncated
## Expected Behavior
No segfault.

## Current Behavior
segfault.

## Possible Solution
It would be great to add some check which would detect truncation of
source file between generating the file list and actually creating checksums
and opening the file for transfer in send_files function in sender.c module.

## Steps to Reproduce
1. $ mkdir dir1; mkdir dir2;
2. $ truncate -s 2760M dir1/file; truncate -s 2759M dir2/file;
3. $ gdb /usr/bin/rsync < script.txt

Attaching script.txt, because this issue is time dependent. 
[script.txt](https://github.com/WayneD/rsync/files/5167233/script.txt)

Full invocation of problematic command is:
`$ rsync --append ./dir1/file ./dir2`

## Context (Environment)
User encountered this issue when trying to sync 2 files and the source
file got truncated to 0 bytes during this operation. I can reproduce this
with the most recent version of rsync on Fedora.

## Detailed Description
The way i understand this is, that the file passes to cur_flist because
it is bigger than the destination, then when it gets truncated, do_fstat
function sets st.st_size to 0 and thus, mbuf is now NULL, because of
condition on sender.c:382 `if (st.st_size)`
After that, match_sums function gets called with buf == NULL and len == 0
Because of that, matched function gets called on match.c:409
and there overflow happens while computing variable n.
`int32 n = (int32)(offset - last_match); /* max value: block_size (int32) */`
offset == 0 and last_match is too big for int32 type. Value of n is now
positive because of the overflow.
On match.c:123
`	for (j = 0; j < n; j += CHUNK_SIZE) {`
j is now smaller than n and map_ptr function gets called with map == NULL

Finally here we are on fileio.c:247 where map gets dereferenced and so we
end up with SIGSEGV.

**Thanks for your time and help.**

TomasKorbar:
Hi,
perhaps a simple check to find out if the file was truncated between receiving the attributes
and stat is enough to determine if it safe to contine. This if causes rsync to skip files which
were truncated like that and continues with the rest of the file list.
Fix #90

### DIFF
--- a/sender.c
+++ b/sender.c
@@ -362,6 +362,13 @@ void send_files(int f_in, int f_out)
 			exit_cleanup(RERR_FILEIO);
 		}
 
+		if (append_mode > 0 && st.st_size < file->len32) {
+			rprintf(FWARNING, "skipping file %s because it got truncated\n", full_fname(fname));
+			free_sums(s);
+			close(fd);
+			continue;
+		}
+
 		if (st.st_size) {
 			int32 read_size = MAX(s->blength * 3, MAX_MAP_SIZE);
 			mbuf = map_file(fd, st.st_size, read_size, s->blength);


WayneD:
Thanks for the suggested patch!  I've fixed this in a similar way, but with various changes.

PR #102, opened 2020-09-20
optimize avx2 codepath

Shark64:
Hi, i was playing around with the avx2 enabled checksum function and noticed that you could reduce the number of shuffles needed if you interleave the data inside the ymm registers. Basically we do two 128bit lanes in parallel and merge the result at the end, instead of using the full 256bit registers as a single vector. The code is based on the SSSE3 one, with small tweaks to reduce the number of immediate that need to be loaded from memory.
I've run the embedded benchmark and the AVX2 codepath went from ~12GB/s to ~14.8GB/s on and Intel 6700k. 
The checksum in the benchmark matches that of the other implementations so hopefully the algorithm is equivalent, but i've only tested it on linux-x86_64 with gcc 10.


WayneD:
Thanks!

PR #103, opened 2020-09-21
Mention -a, --archive does not include atime, and thereby hint it is possible

richardkmichael:
Rsync gaining `-U, --atimes` in 3.2.0 is a recent but helpful addition; publicize it by mentioning it is *not* included in `-a, --archive`.

I think it would be reasonable to expect an archive to respect atimes, but changing `-a` itself is likely a breaking change (or at least perhaps a surprising one) for long-time rsync users.


WayneD:
Thanks for the patch!  I've committed a slightly improved version that also mentions `-N` and expands the long text a bit.

PR #137, opened 2021-01-15
Add link to manpage on GitHub

claell:
Might not be wanted, since there are also other options to display the manpage.


WayneD:
Thanks for the patch.  I've opted to tweak your idea to link to the html version on rsync.org since it renders the defined list properly.  I also like to keep the README readable as a plain-text file, so I've been using the [num] reference idiom for links.

PR #160, opened 2021-02-28
AVX2: delay line crossing add until the end of loop

Shark64:
On all x86-64 CPU with AVX2 the instructions that extract/insert across the 128bit "lane" are slower than the "in lane" ones (for example, on skylake-s 3 cycles vs 1 of latency). To avoid using them inside the checksum loop we can keep two partial sum in each of  the two 128 bit lane and delay the final sum after the loop is over. On my i7-6700 under Linux the simd benchmark went from ~16Gb/sec to ~19--20 Gb/sec. I've not tested on an AMD cpu but should be faster there too.
I've also increased the prefetch distance to 6 cacheline because 2 cachelines are almost always prefetched by the hardware automatically.
The ./bench output gives the same checksum as the other functions, so hopefully there're no new bugs  ;)


Shark64:
I've noticed that the FreeBSD check has failed, seems like an issue with the m4 version, nothing to do with the patch right?

Shark64:
I've noticed that you've updated the Cirrus CI config file, so i've merged your changes into my repository. Everything seems to be working now, thanks :)

Shark64:
With the last changes both halves of the checksum are kept as partial sums in the vector registers and reduced at the end of the loop. On my machine i get ~30GB/s on average, same checksum as the other functions so everything should work as expected.

Shark64:
@WayneD do you prefer if i close this merge request and open another one with the same code but without all the intermediate commits logged? Works for me either way, thanks

WayneD:
No need to worry about intermediate fixes in a merge request, as I always squash merged changes.  I hope to check this out soon. Thanks!

Shark64:
Ok nice :). 
I had some time to play around with the code emitted by GCC and it doesn't look very good to be honest. I've written an assembly version of get_checksum1_avx2() on my local machine that is both smaller and a bit faster than what GCC emits from the intrinsics, no matter how i tried to play around with the compiler flags. I edited the intrinsics because i was trying to avoid touching the Makefiles, but i've finally give up and tried to add that function to the build system. Seems to works on my machine but i'm more of a low-level guy so don't feel 100% sure messing around with automake stuff. 
If you have time to review them i'd rather close this and send you the pure assembly pull-request, let me know what you think of that 😄

WayneD:
I'm fine with the change if that's what you'd like to do -- feel free.

PR #161, opened 2021-03-03
FreeBSD 12.2

fkorotkov:
12.1 Reached End-Of-Life https://www.freebsd.org/releases/


WayneD:
Thanks for the suggested fix!  I fiddled the CI file before noticing your push request, but ended up in the same place.

PR #174, opened 2021-04-09
x86-64 AVX2 assemby implemenation of get_checksum1()

Shark64:
Here's the assembly version of get_checksum1 optimizated for AVX2 instructions. On my Core i7-6700 the benchmark reaches ~45Gb/sec and it's quite smaller than the gcc+instrinsics version. 
Should i add a copyright comment at the beginning of the file? I was not sure about that. 
Thanks


Shark64:
I've added the #ifdef __APPLE__ like the ones in the md5 assembly implementation, to see if those were enough to fix the build failure under MacOS. Unfortunately it seems that there's a problem with the IO transfer during the tests, but i'm not sure if the problem is with my patch or somewhere else and i don't have an Apple computer at hand to try debugging it. Anything you wanna me to do to help with the build failure?

Shark64:
@WayneD i've double checked my patch but can't find any errors during testing under Linux. I don't have access to a MacOS system so i'm not sure if it's a problem with my pull request or of the build system under Apple. Can you please check it and let me know if there's anything i can do to help fix the build failure? thanks :)

PR #178, opened 2021-04-20
man: document possible format strings

mpontillo:
Hi all, I'd like to get some feedback on this potential change.

Today I was trying to figure out what formatting options `rsync` accepts via the manual page, and I didn't find what I was looking for. So instead I tried reading the code.

For the benefit of others who might run into this in the future, I decided to try documenting them in `rsync.1.md`.

Since I don't know the `rsync` code well, some of my descriptions might be wrong (or might need more detail on the subtleties). For example, I'm not sure how to properly describe `%b` or `%c` (there are some subtle differences in the logic depending on which is chosen, the `ITEM_TRANSFER`, and I wasn't sure what was meant by `initial_data_written` and `initial_data_read`). And I assumed from some of the context in the code that `%m` (the module name) had something to do with syslog, but I'm not positive about that (and I was further confused by `%P`, the "full module path", but my casual reading of the code didn't help me understand exactly what the "module" was for.

I also noticed that there was inconsistent usage of `FMT` vs. `FORMAT` in the manual page, but it seems to refer to the same functionality. So I tried to correct that, too.

Lastly, I don't have the tooling installed to generate the manual page from the Markdown file, so if this is something that would be good to move forward with, I'll look into installing that.

### DIFF
--- a/rsync.1.md
+++ b/rsync.1.md
@@ -454,7 +454,7 @@ detailed description below for a complete description.
 --remote-option=OPT, -M  send OPTION to the remote side only
 --out-format=FORMAT      output updates using the specified FORMAT
 --log-file=FILE          log what we're doing to the specified FILE
---log-file-format=FMT    log updates using the specified FMT
+--log-file-format=FORMAT log updates using the specified FORMAT
 --password-file=FILE     read daemon-access password from FILE
 --early-input=FILE       use FILE for daemon's early exec input
 --list-only              list the files instead of copying them
@@ -487,7 +487,7 @@ accepted:
 --no-detach              do not detach from the parent
 --port=PORT              listen on alternate port number
 --log-file=FILE          override the "log file" setting
---log-file-format=FMT    override the "log format" setting
+--log-file-format=FORMAT override the "log format" setting
 --sockopts=OPTIONS       specify custom TCP options
 --verbose, -v            increase verbosity
 --ipv4, -4               prefer IPv4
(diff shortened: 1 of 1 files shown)


rsync.1.md
```diff
@@ -4053,6 +4053,58 @@ If you are having trouble debugging filter patterns, then try specifying the
...
+
+0.  `%c` Total data read
+
+0.  `%c` Checksum
+
+0.  `%i` Itemized changes string (see description of `--itemize-changes`).
+
+# DAEMON FORMAT STRINGS
+
+The following forrmat strings are supported for options accepting a `FORMAT`
```
mamercad, line 4101: Typo: `format`

rsync.1.md
```diff
@@ -4053,6 +4053,58 @@ If you are having trouble debugging filter patterns, then try specifying the
 `-vv` option.  At this level of verbosity rsync will show why each individual
 file is included or excluded.
 
+# FORMAT STRINGS
+
+The following forrmat strings are supported for options accepting a `FORMAT`
```
mamercad, line 4058: Typo: `format`

mamercad (commented):
I like this; it would be nice for some of the format options to provide an example of what they look like, e.g., what does "`%b` Total data written" look like typically?

rsync.1.md
mpontillo: Fixed; thanks!

mpontillo:
I agree that some examples would be nice. When I get a chance, I'll see if I can test some of this to check my work and provide examples. (I hesitated to add too much text yet, since I need to set up the `md2man` tool and take a look at how the output gets rendered in the manual.)

WayneD:
The man page for --log-file-format=FORMAT mentions that the strings are documented in the rsyncd.conf man page. I don't see any need to duplicate them in both man pages.

Issue #177, opened 2021-04-17
--time-limit hangs instead of exiting


ppaeps:
Since upgrading to rsync 3.2.3, it looks like rsync doesn't exit when hitting the specified `--time-limit=minutes`.  It stops transferring data and reports that the time limit has been reached, but the process never exits.

```
stopping at requested limit
rsync error: timeout in data send/receive (code 30) at io.c(794)
[receiver=3.2.3]
```

We're running:
```
# rsync --version
rsync  version 3.2.3  protocol version 31
Copyright (C) 1996-2020 by Andrew Tridgell, Wayne Davison, and others.
Web site: https://rsync.samba.org/
Capabilities:
    64-bit files, 64-bit inums, 64-bit timestamps, 64-bit long ints,
    socketpairs, hardlinks, hardlink-specials, symlinks, IPv6, atimes,
    batchfiles, inplace, append, ACLs, xattrs, optional protect-args,
    no iconv, symtimes, no prealloc, stop-at, no crtimes
Optimizations:
    no SIMD, no asm, openssl-crypto
Checksum list:
    xxh128 xxh3 xxh64 (xxhash) md5 md4 none
Compress list:
    zstd lz4 zlibx zlib none
```

It's simple enough to reproduce.  Please let me know if you need more data though.

rosorio:
After analysis, it seems that time-limit options only applies to receiver, I submit a path #179 to evaluate the delay at every io option.

PR #179, opened 2021-04-20
Time limit options applies to send and receive transfers

rosorio:
stop-at, stop-after and time-limit options must be transfer
direction agnostic, and must apply to rsync in receiver
or sender mode.

### DIFF
--- a/io.c
+++ b/io.c
@@ -787,18 +787,19 @@ static char *perform_io(size_t needed, int flags)
 			if (msgs2stderr == 1 && DEBUG_GTE(IO, 2))
 				rprintf(FINFO, "[%s] recv=%ld\n", who_am_i(), (long)n);
 
-			if (io_timeout || stop_at_utime) {
+			if (io_timeout) {
 				last_io_in = time(NULL);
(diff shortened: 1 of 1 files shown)


ppaeps:
This patch appears to work.  Previously, rsync 3.2.3 would simply hang, as reported in #177, with this patch, it exits as expected:

```
stopping at requested limit
rsync error: timeout in data send/receive (code 30) at io.c(802) [receiver=3.2.3]
stopping at requested limit
rsync error: timeout in data send/receive (code 30) at io.c(802) [generator=3.2.3]
```

For completeness, this is what used to happen with 3.1.3:

```
run-time limit exceeded
rsync error: timeout in data send/receive (code 30) at io.c(181) [generator=3.1.3]
rsync error: received SIGUSR1 (code 19) at main.c(1441) [receiver=3.1.3]
```

Thank you very much!

rosorio:
Hi @WayneD , do you plan to integrate this fix in a future release ?

ppaeps:
Gently prodding here.  We've been using this patch in production without a problem since April.

WayneD:
Thanks!  Sorry for the huge delay.

PR #198, opened 2021-07-03
Add independent update mode for symbolic links

twojstaryzdomu:
This PR enhances rsync with an independent update mode for symbolic links. Its main purpose is to skip symlinks that are newer on the receiver. The mode is enable via the '--update-links' switch. To extend the behaviour of this option to directories, an addition switch '--allow-link-update-dir' may be enabled. In this mode, less recent directory on the destination may be replaced with a more recent symbolic link from the source.
The PR includes a sanity test to verify the impact of the proposed modes on a set of files (relative dangling absolute unsafe devlink dirlink) under different circumstances.

### DIFF
--- a/flist.c
+++ b/flist.c
@@ -62,6 +62,7 @@ extern int implied_dirs;
 extern int ignore_perishable;
 extern int non_perishable_cnt;
 extern int prune_empty_dirs;
+extern int update_links;
 extern int copy_links;
 extern int copy_unsafe_links;
 extern int protocol_version;
@@ -231,10 +232,15 @@ static int readlink_stat(const char *path, STRUCT_STAT *stp, char *linkbuf)
 int link_stat(const char *path, STRUCT_STAT *stp, int follow_dirlinks)
 {
 #ifdef SUPPORT_LINKS
-	if (copy_links)
+	if (copy_links && update_links == 0)
 		return x_stat(path, stp, NULL);
 	if (x_lstat(path, stp, NULL) < 0)
 		return -1;
+	if (update_links && S_ISLNK(stp->st_mode)) {
+		STRUCT_STAT st;
+		if (x_stat(path, &st, NULL) == 0 && !S_ISDIR(st.st_mode))
+			return x_lstat(path, stp, NULL);
+	}
 	if (follow_dirlinks && S_ISLNK(stp->st_mode)) {
 		STRUCT_STAT st;
 		if (x_stat(path, &st, NULL) == 0 && S_ISDIR(st.st_mode))
--- a/generator.c
+++ b/generator.c
@@ -55,6 +55,9 @@ extern int ignore_errors;
 extern int remove_source_files;
 extern int delay_updates;
 extern int update_only;
+extern int update_links;
+extern int copy_links;
+extern int allow_link_update_dir;
 extern int human_readable;
 extern int ignore_existing;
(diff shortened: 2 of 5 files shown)


twojstaryzdomu:
The test included in the PR needs further work to run on FreeBSD apparently.

WayneD:
I'll note that tweaking `.github/workflows/build.yml` to include `coreutils` on the `brew install` line should hopefully be enough to fix the test error on macOS.

twojstaryzdomu (commented):
The redundant calls to the macro were intentional, commit reverted.

twojstaryzdomu:
> I'll note that tweaking `.github/workflows/build.yml` to include `coreutils` on the `brew install` line should hopefully be enough to fix the test error on macOS.

Noted and added it in commit `c3437c1`. The workflow change now awaits approval. #

flist.c
```diff
@@ -231,10 +232,15 @@ static int readlink_stat(const char *path, STRUCT_STAT *stp, char *linkbuf)
 int link_stat(const char *path, STRUCT_STAT *stp, int follow_dirlinks)
 {
 #ifdef SUPPORT_LINKS
-	if (copy_links)
+	if (copy_links && update_links == 0)
```
WayneD, line 235: If they asked for --copy-links they should get --copy-links. Remember that it's only the sending side that sees this flag set, not the receiving side, so combining --copy-links with --update-links would mean that the sender doesn't send any symlinks, and the receiver could deal with any existing symlinks in the destination using the new option. (Thus, drop this change.)

flist.c
```diff
@@ -231,10 +232,15 @@ static int readlink_stat(const char *path, STRUCT_STAT *stp, char *linkbuf)
...
 		return x_stat(path, stp, NULL);
 	if (x_lstat(path, stp, NULL) < 0)
 		return -1;
+	if (update_links && S_ISLNK(stp->st_mode)) {
+		STRUCT_STAT st;
+		if (x_stat(path, &st, NULL) == 0 && !S_ISDIR(st.st_mode))
+			return x_lstat(path, stp, NULL);
+	}
```
WayneD, line 243: This hunk baffles me. It does a stat of the referent file, and if it's a non-dir it re-stats the symlink's stat that is already present in stp and returns. But if it is not a dir, the following code also returns the same stat data that is already in stp. So, it seems like this can be dropped as well.

flist.c
twojstaryzdomu: These two modes appear mutually exclusive from a user standpoint. Either every symlink is transformed into referent (`--copy-links`) or copied across if newer (`--update-links`). 

> so combining --copy-links with --update-links would mean that the sender doesn't send any symlink

From testing the PR'd code briefly now, when both modes are enabled, `--update-links` trumps `--copy-links`, by virtue of a subsequent condition at `if (update_links && S_ISLNK(stp->st_mode))`.

Added #274 to discuss how to handle symlink mode handling precedence.

flist.c
twojstaryzdomu: Precisely, if the symlink points at a non-directory referent, return the symlink's stat rather than the referent or fall through. It's only apparent looking at the surrounding code why the x_lstat call is redundant. Committed as `aed6235`, works on assumption lstat or lstat64 is guaranteed to return either 0 or -1 on all platforms.

Issue #192, opened 2021-06-14
Possible issue with --delay-updates and stale partial data


foutrelis:
I have noticed the following behavior on our Arch Linux mirrors which leaves stale partial data behind and causes those files to never be updated.

Below is a minimal reproduction case of the behavior I'm seeing:

```
mkdir source dest
echo 1 >source/test
rsync -av --delay-updates source/ dest/

mkdir dest/.~tmp~
echo 2 >dest/.~tmp~/test
echo 3 >source/test
rsync -av --delay-updates --modify-window=-1 source/ dest/
```

Repeating the last rsync command above will always show `test` as being transferred. It appears that `dest/.~tmp~/test` gets updated but never renamed to `dest/test`.

As to why the file in the partial directory exists in the first place, it seems to be left over from an earlier transfer that failed verification (seems to happen with files that get updated in place, not with other repository files with unique filenames). At least that's what I gather from [our logs](https://gitlab.archlinux.org/archlinux/infrastructure/-/issues/227):

```
Oct 17 18:38:01 dragon.archlinux.org syncrepo[2017609]: WARNING: community/os/x86_64/community.files.tar.gz failed verification -- update discarded (will try again).
Oct 18 12:41:04 dragon.archlinux.org syncrepo[3140845]: WARNING: community/os/x86_64/community.files.tar.gz failed verification -- update put into partial-dir (will try again).
Oct 19 17:17:03 dragon.archlinux.org syncrepo[358950]: WARNING: lastsync failed verification -- update discarded (will try again).
Oct 21 22:41:03 dragon.archlinux.org syncrepo[826587]: WARNING: lastsync failed verification -- update put into partial-dir (will try again).
```

Apologies if this isn't a bug and we're just using rsync incorrectly; in which case, pointers about what we should change would be greatly appreciated! (The real command for mirror syncing is something like `rsync -rlptH --safe-links --delete-after 
--delay-updates --temp-dir=/srv/syncrepo-tmp rsync://rsync.archlinux.org/ftp_tier1 /srv/ftp`.)

foutrelis:
[...]

PR #204, opened 2021-07-21
Add test and fix regression for --delay-updates (#192)

ncopa:
to test https://github.com/WayneD/rsync/issues/192


WayneD:
Looks good to me -- much appreciated!

PR #220, opened 2021-09-08
Fix spelling of 'nanosecond' ...

zgpmax:
... to make it easier to find when nanosecond support was added (3.1.0 / 28 Sep 2013)


zgpmax:
Thank-you chief 👍😄

PR #238, opened 2021-10-10
Stop rrsync escaping restricted folder, by forcing --munge-links

sourcejedi:
Fixes: https://bugzilla.samba.org/show_bug.cgi?id=11879

> fb102email-sambabugzilla 2016-04-28 15:52:40 UTC
>
> It is possible to escape rrsync restricted folder by syncing (using
> rsync -a ...) a symbolic link to the parent folder and then syncing with
> this symbolic link. [...]
>
> g.parrondo 2016-05-26 16:12:06 UTC
>
> Adding '--safe-links' or '--munge-links' on server side should fix this.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -36,6 +36,8 @@
  - Avoid a weird failure if you run a local copy with a (useless) `--rsh`
    option that contains a `V`.
 
+ - Stop rrsync escaping restricted folder, by forcing --munge-links.
+
 ### ENHANCEMENTS:
 
  - Use openssl's `-verify_hostname` option in the rsync-ssl script.
--- a/support/rrsync
+++ b/support/rrsync
@@ -13,6 +13,7 @@ use File::Glob ':glob';
 # of options if you want to disable any options that rsync accepts.
 use constant RSYNC => '/usr/bin/rsync';
 use constant LOGFILE => 'rrsync.log';
+use constant FORCE_OPT => ('--munge-links');
 
 my $Usage = <<EOM;
 Use 'command="$0 [-ro|-wo] SUBDIR"'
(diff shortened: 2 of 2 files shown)


WayneD:
This is a good idea to have this as the default.  I've implemented it in an optional manner that allows the ssh config to use an option to disable the munging (perhaps they trust their users or if they just can't allow munging).

sourcejedi:
Hurrah! Thanks Wayne, that makes sense.

WayneD:
I'll note that I went on to re-implement rrsync as a python3 script and added improved arg checking (it validates that the de-symlinked args do not end up outside the restricted dir) while also rejecting the use of `-L` (`--copy-links`), at which point I no longer think we need to force the munging of symlinks. I have therefore turned the munging off by default and changed the just-added `-no-munge` rrsync option to a `-munge` rrsync option.

sourcejedi:
Thanks for the update.

As I understood it, the new solution requires consideration of [TOCTOU](https://en.wikipedia.org/wiki/TOCTOU).  The bugzilla mentions a similar sounding solution, but [using a lock to avoid TOCTOU](https://bugzilla.samba.org/show_bug.cgi?id=11879#c2). Is that what you implemented?

Personally, I would use the `-munge` option. Because of this being one of the solutions [documented in the security advisory](https://rsync.samba.org/security.html#s3_0_0) for years now.

WayneD:
Yes, you have understood correctly.  I've gone ahead and added a default single-use lock to the latest script, similar to what was suggested in bugzilla.  You can avoid it using `-no-lock` (presumably combined with your preferred `-munge`).

PR #241, opened 2021-10-14
Linux: Handle protected_regular in inplace writes

fjl5:
Linux kernel 4.19 and systemd 241 introduced a new filesystem protection feature called "protected_regular", see https://www.kernel.org/doc/Documentation/sysctl/fs.txt:

> protected_regular:
> 
> This protection is similar to protected_fifos, but it
> avoids writes to an attacker-controlled regular file, where a program
> expected to create one.
> 
> When set to "0", writing to regular files is unrestricted.
> 
> When set to "1" don't allow O_CREAT open on regular files that we
> don't own in world writable sticky directories, unless they are
> owned by the owner of the directory.
> 
> When set to "2" it also applies to group writable sticky directories.

To test if the feature is turned on, you can use
```
$ sudo sysctl fs.protected_regular
fs.protected_regular = 2
```

This may end up in `rsync --inplace` failing to write to some files, even if run as root. A simple `cp -a` succeeds, though, as rsync uses the `O_CREAT` option, while cp does not. This change works around the issue by detecting the failures and trying again without `O_CREAT`. The idea is that these types of failures are rare, so we don't want to change the default logic.

Added a simple test case. As we need chown to trigger the failure, this test is skipped if not running as root:
```
SKIP    protected-regular (Can't chown (probably need root))
```

Without the patch, `sudo make test` shows:
```
----- protected-regular log follows
Testing for symlinks using 'test -h'
Contents of /home/fjl/rs/rsync.orig/testtmp/protected-regular/files:
insgesamt 32
drwxrwxrwt 2 root root 4096 Okt 14 10:29 .
drwxr-xr-x 3 root root 4096 Okt 14 10:29 ..
-rw-r--r-- 1 5001 root    1 Okt 14 10:29 dst
-rw-r--r-- 1 root root    7 Okt 14 10:29 src
rsync: [receiver] open "/home/fjl/rs/rsync.orig/testtmp/protected-regular/files/dst" failed: Permission denied (13)
rsync error: some files/attrs were not transferred (see previous errors) (code 23) at main.c(1334) [sender=v3.2.3-112-g0f87eafa]

----- protected-regular log ends
FAIL    protected-regular
```

With the patch, we have:
```
PASS    protected-regular
```


fjl5:
Perfect, thank you!

PR #250, opened 2021-11-01
shell: test -a|o is not POSIX

concatime:
I faced `test: too many arguments` when building using [sbase](https://core.suckless.org/sbase/).
This is due to a non-POSIX syntax `test ... -a ...` and `test … -o …`.

> The XSI extensions specifying the -a and -o binary primaries and the '(' and ')' operators have been marked obsolescent.

https://pubs.opengroup.org/onlinepubs/9699919799/utilities/test.html


WayneD:
Thanks!

PR #259, opened 2021-11-25
disabling asm optimization [Adva] 

wikjas:
Hello,

With my PR i would like to start a discussion about asm optimization for md5 checksum, so i am aware that my changes are not proper to merge with master.

I did some tryouts rsyncing the same files for two diffrent platforms, for the first tryout i used rsync build with --disable-asm option and for the other one the same files were transfered with rsync build with asm optimization. I runned rsync with --chechsum-choice=md5 option and measured the time rsync needed to transfer files. The results below:

file1	 4,419,818,083 bytes,  no asm:    3m48.141s,   asm:    4m24.504s

file2	 224,763,067 bytes,     no asm:    0m12.875s,   asm:    0m22.460s

file3	352,407,655 bytes,      no asm:    0m19.237s,   asm:    0m28.148s

As You can see, using asm optimization extended the time needed for file transfer. Even for a quick test like this the diffrences are noticeable. Maybe there is no need to use this optimization at all, especially when it works only for x86_64 arch, or maybe i missed something here.


WayneD:
I did some timings of the asm code, and find that it is about 3 times faster than the C code.  The code is not used if the openssl MD5 code is being used (and is about the same speed as that code).

PR #265, opened 2021-12-26
Prefer mallinfo2() if available

jmakovicka:


WayneD:
Thanks for the suggestion and the patch.  I decided to make the use of %zd conditional on a C99 check and then made use of %zd in various debug prints in io.c as well as the allocation stats that you modified.

PR #273, opened 2022-01-07
Quotation for usermap and groupmap arguments

### LINKED ISSUE #272: chown leads to "No matches for wildcard" in combination with fish shell
When using the `--chown` flag with rsync on a target server that has fish shell as the default for the user, rsync errors out with a "No matches for wildcard". fish always eagerly tries to expand wildcards, which is not possible in this case.

```
$ rsync -avz --chown root:root . user@example.com.com:/srv/html/example.com/.
root@example.com's password: 
fish: No matches for wildcard “--usermap=*:root”. See `help expand`.
rsync --server -vlogDtprze.iLsfxCIvu --usermap=*:root --groupmap=*:root . /srv/html/example.com/.
                                     ^
rsync error: error in rsync protocol data stream (code 12) at io.c(228) [sender=3.2.3]
```

rsync version: `rsync  version 3.2.3  protocol version 31`

Wouldn't it be possible to escape the asterisk or use quotes in this case?

brejoc:
Since wildcards are being used as parameters, shells like Fish try to expand this and fail. The single quotes prevent them from doing so and still work with Bash.

Fixes #272

### DIFF
--- a/options.c
+++ b/options.c
@@ -2783,13 +2783,13 @@ void server_options(char **args, int *argc_p)
 
 	if (am_sender) {
 		if (usermap) {
-			if (asprintf(&arg, "--usermap=%s", usermap) < 0)
+			if (asprintf(&arg, "--usermap='%s'", usermap) < 0)
 				goto oom;
 			args[ac++] = arg;
 		}
 
 		if (groupmap) {
-			if (asprintf(&arg, "--groupmap=%s", groupmap) < 0)
+			if (asprintf(&arg, "--groupmap='%s'", groupmap) < 0)
 				goto oom;
 			args[ac++] = arg;
 		}


brejoc:
Since I know next to nothing about C, this should just be seen as an educated guess. But at least it seems to be working with Fish and Bash.

WayneD:
This would break `--protect-args` (since the quotes would become a part of the actual arg).  Also, it only protects the `*` through the first ssh call, not on the receiving shell.  So, when protect_args is not set, the code could possibly send it surrounded in both single and double quotes, but it might be better to turn each `*` into `\\\*` since that gives the shell that is running ssh the `\\\*` and then the remote shell gets `\*` (which it turns into `*`).  I'm fiddling around with  my own patch for this.  Thanks, though!

PR #291, opened 2022-02-14
Correct units in --stats

wisniew:


tridge:
@WayneD I'd be interested in your view, but I don't think this is a worthwhile change, I'm sure some people have perl/python scripts to parse this info. We could add an option to give machine readable output if needed

PR #295, opened 2022-03-06
Add missing include to largefile configure check

jmroot:
The configure check for broken largefile support calls strcpy but was not including string.h.


WayneD:
Thanks for pointing that out.  I've tweaked configure in a slightly different way, so I'll close this.

PR #306, opened 2022-04-05
Fix for CVE-2018-25032 in zlib

mruprich:
A CVE was found in zlib that was introduced in zlib-1.2.2.2 and affects every version released after that including the bundled version in rsync. I've used the original patch with one or two small changes where it would not fit the bundled version.

Adding the zlib upstream fix here, the detailed explanation is in the commit - https://github.com/madler/zlib/commit/5c44459c3b28a9bd3283aaceab7c615f8020c531


mruprich:
Great, thanks :)

PR #308, opened 2022-04-18
rsync.1.md: fix "specifed" typo

samueloph:
Caught this issue when packaging the latest release for Debian.

Thanks,


WayneD:
Good catch.  I was making some manpage improvements, so I tweaked the spelling in that commit. Thanks for pointing out the issue, though!

PR #311, opened 2022-04-25
Use SOURCE_DATE_EPOCH on manpage generation (reproducible build)

samueloph:
Whilst working on the Reproducible Builds effort [0] we noticed that
rsync could not be built reproducibly.

This was because the manpage generation used the current date for the
manual page header.

A patch is attached that will use SOURCE_DATE_EPOCH if available.

[0] https://reproducible-builds.org/

### DIFF
--- a/md-convert
+++ b/md-convert
@@ -241,7 +241,7 @@ def find_man_substitutions():
                 if var == 'srcdir':
                     break
 
-    env_subs['date'] = time.strftime('%d %b %Y', time.localtime(mtime))
+    env_subs['date'] = time.strftime('%d %b %Y', time.gmtime(int(os.environ.get('SOURCE_DATE_EPOCH', time.time()))))
 
 
 def html_via_commonmark(txt):


WayneD:
after git clone, run `support/git-set-file-times`

WayneD:
No, it is not getting today's date. It is getting the date from the last-modified time.  If you have a git checkout, that time is the date of the last commit.  If you are using an official release tar, the manpage date is taken from the date of the version.h file (which will be the date the tar was created).  So, for normal source-grabbing methods, you should end up with a consistent build.

If you are doing something weird, such as cloning the repo and removing the .git dir, consider instead using a `git archive --git-dir=FOO TAG | tar xf - -C SUBDIR` command, since that makes the date of all the extracted files the date of the last commit.  Or, if you must use the .git-removing method, run `support/git-set-file-times` right after the clone (before the rm).  If you're doing something else, feel free to touch the version.h file with the time you desire.

samueloph:
@WayneD That's right, the date used in the manpage is not the current date at build time, but the issue still persists.

The date used is mtime of version.h, but the command used to convert from the timestamp is python's [time.localtime()](https://docs.python.org/3/library/time.html#time.localtime), which will return different dates depending on the TZ, you can see this happening here:
https://tests.reproducible-builds.org/debian/rb-pkg/bookworm/amd64/diffoscope-results/rsync.html

Alternatively, and probably better than the initial PR, this can be fixed by using [time.gmtime()](https://docs.python.org/3/library/time.html#time.gmtime) instead.

I have updated the PR to fix this by `s/localtime/gmtime/` instead, what do you think?

I could not update/reopen the original PR after force-pushing to my fork, Github lost track of it so I ended up having to create a new one.

New PR:
https://github.com/WayneD/rsync/pull/316

Issue #318, opened 2022-05-07
[v3.2.4] absolute path resolved as relative to home directory when double-quoted in single quotes


twojstaryzdomu:
Rsync release 3.2.4 appears to be resolving a double-quoted absolute path enclosed within a single quoted destination as a path relative to a user's home directory.

Quoting the path element should not be affect whether that path is absolute or not.

The overarching single quoting is a result of shell quoting within a ksh shell script. This is done to ensure the destination is properly quoted (to address possible spaces).

The exact command statement from the shell script is as follows:

`rsync "${RSYNC_OPTS[@]}" "${SOURCE}/" "${TARGET}"`

Where `TARGET` is a variable that contains a double quoted component.

The case is equivalent to assigning a variable containing double quotes to another variable, which is later quoted. 

The quoting problem may be reproduced at an interactive shell prompt (bash):
```
# rsync -vvv '-essh -o StrictHostKeyChecking=no -o CheckHostIP=no' /etc/passwd 'username@hostname:"/tmp"'
opening connection using: ssh -o StrictHostKeyChecking=no -o CheckHostIP=no -l username hostname rsync --server -vvve.LsfxCIvu . "\"/tmp\""  (13 args)
[sender] make_file(passwd,*,0)
send_file_list done
send_files starting
server_recv(2) starting pid=26808
rsync: [Receiver] change_dir#3 "/home/username/"" failed: No such file or directory (2)
rsync error: errors selecting input/output files, dirs (code 3) at main.c(818) [Receiver=3.2.4]
recv_file_name(passwd)
received 1 names
recv_file_list done
get_local_name count=1 "/tmp"
[Receiver] _exit_cleanup(code=3, file=main.c, line=818): about to call exit(3)
[sender] _exit_cleanup(code=3, file=io.c, line=1667): about to call exit(3)
```
The problem isn't seen with rsync release 3.2.3:
```
# rsync --version
rsync  version v3.2.3  protocol version 31
# rsync -vvv '-essh -o StrictHostKeyChecking=no -o CheckHostIP=no' /etc/passwd 'username@hostname:"/tmp"'
opening connection using: ssh -o StrictHostKeyChecking=no -o CheckHostIP=no -l username hostname rsync --server -vvve.LsfxCIvu . ""/tmp""  (13 args)
[sender] make_file(passwd,*,0)
send_file_list done
send_files starting
server_recv(2) starting pid=30628
recv_file_name(passwd)
received 1 names
recv_file_list done
[...]

PR #320, opened 2022-05-14
Define the 'really old versions' for '--old-args' for clarity

twojstaryzdomu:
It's not clear what 'Really old versions' is supposed to mean. With the `--old-args` being the default in 3.2.3, users may be scratching their heads why 'really old versions' refers to a version only one minor release behind.

This PR is supposed to related to the loss of backwards compatibility for parameter parsing as detailed in #318.

### DIFF
--- a/rsync.1.md
+++ b/rsync.1.md
@@ -182,7 +182,7 @@ spaces from the local shell but not also from the remote shell:
 
 >     rsync -aiv host:'a simple file.pdf' /dest/
 
-Really old versions of rsync only allowed specifying one remote-source arg, so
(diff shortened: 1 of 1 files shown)


WayneD:
Not true in any way shape or form.

WayneD:
Clarified in the latest manpage: https://download.samba.org/pub/rsync/rsync.1#ADVANCED_USAGE

twojstaryzdomu:
> Clarified in the latest manpage: https://download.samba.org/pub/rsync/rsync.1#ADVANCED_USAGE

Appreaciate it, it's now clear what a user must do to maintain backwards compatibility.

PR #369, opened 2022-09-05
Use string length diff heuristic to skip Levenshtein Algo

PhirePhly:
When using the --fuzzy option to try and find close matches locally, the edit distance algorithm used is O(N^2), which can get painful on CPU constrained systems when working in folders with tens of thousands of files in it.

The lower bound on the calculated Levenshtein distance is the difference of the two strings being compared, so if that difference is larger than the current best match, the calculation of the exact edit distance between the two strings can be skipped.

Testing on the OpenSUSE package repo has shown a 50% reduction in the CPU time required to plan the rsync transaction. Some of the OpenSUSE mirrors I operate have been seeing 2-4 hours of CPU time on an rsync run when there's a large rebuild in the folder.

Performing a smaller test where I deleted 1000 random files from one folder:

```
kenneth@node1:~/tmp$ time ~/src/rsync.optim -aSHy --dry-run rsync://codingflyboy.mm.fcix.net/opensuse/ ./opensuse/           
real    7m11.766s
user    7m8.904s
sys     0m1.258s
kenneth@node1:~/tmp$ time ~/src/rsync.mainline -aSHy --dry-run rsync://codingflyboy.mm.fcix.net/opensuse/ ./opensuse/

real    12m16.094s
user    12m12.873s
sys     0m1.240s
```

### DIFF
--- a/generator.c
+++ b/generator.c
@@ -875,9 +875,12 @@ static struct file_struct *find_fuzzy(struct file_struct *file, struct file_list
 			len = strlen(name);
 			suf = find_filename_suffix(name, len, &suf_len);
 
-			dist = fuzzy_distance(name, len, fname, fname_len);
-			/* Add some extra weight to how well the suffixes match. */
-			dist += fuzzy_distance(suf, suf_len, fname_suf, fname_suf_len) * 10;
+			dist = fuzzy_distance(name, len, fname, fname_len, lowest_dist);
+			/* Add some extra weight to how well the suffixes match unless we've already disqualified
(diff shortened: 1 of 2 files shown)


PhirePhly:
I suspect there's a second possible optimization aborting the Levenshtein calculation once we've made it far enough into the nested loops that we know that we've exceeded the `upperbound`, but I'm still trying to wrap my head around how the vector implementation of the algorithm works.

WayneD:
Thanks!

PR #374, opened 2022-09-10
usage.c: Fix typo "Licence"

samueloph:
Spotted by lintian when packaging 3.2.6 for Debian.


WayneD:
Since Tridge is from Australia, that spelling of the word is a callback to the origins of rsync (it is spelled that way there).  I've been inclined to leave it alone, but it is a bit inconsistent with the American spellings in other places in the docs these days.

samueloph:
I wasn't aware that "Licence" could be a proper spelling of the word, thanks for letting me know.

I'll try to report this back to Lintian so it stops complaining about this.

PR #377, opened 2022-09-14
Support spaces in paths when using rrsync

mattcen:
The rrsync wrapper script was using a regex to split SSH_ORIGINAL_COMMAND, which wasn't accounting for filesystem paths containing spaces. This commit switches to using shlex to parse the command line, resolving this issue.

### DIFF
--- a/support/rrsync
+++ b/support/rrsync
@@ -130,7 +130,7 @@ long_opts = {
 
 ### END of options data produced by the cull-options script. ###
 
-import os, sys, re, argparse, glob, socket, time, subprocess
+import os, sys, re, argparse, glob, socket, time, subprocess, shlex
 from argparse import RawTextHelpFormatter
 
 try:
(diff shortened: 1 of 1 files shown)


WayneD:
To be clear: rrsync already supports backslash-escaped spaces, such as those sent by rsync 3.2.4 and beyond.  What it doesn't allow are other types of arg escaping, such as quoted strings.

So, using shlex would be nice, but only if it would return all the backslashes intact.  We need the user's backslashes for passing the string to braceexpand(). Otherwise something like "\{foo,bar\}" will not be protected from brace expansion.

In looking at the shlex functions, it looks like calling `s = shlex.shlex(command_line)` and setting `s.whitespace_split = True` could get us something that we could parse arg-by-arg to build up a command-line.  The parsing would need to distinguish quoted args (and remove the quotes) from non-quoted args where a trailing odd-count of backslashes would combine the current arg + a space with the next arg (which might or might not be quoted).

We'd also need to handle any raised errors from shlex and return a nicer error message than a python stack dump.

PR #390, opened 2022-10-15
introduce mmap as option

Alexponomarev7:
I know that mmap makes many unpleasant effects, but in my cases this option really useful because my pattern of usage guarantees no problem with mmap, I guess it may be quite a popular among 'build from source' users and for benchmarking, of course

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -84,6 +84,15 @@ if test x"$enable_profile" = x"yes"; then
 	CFLAGS="$CFLAGS -pg"
 fi
 
+AC_ARG_ENABLE(mmap,
+    AS_HELP_STRING([--enable-mmap], [enable to omit mmap]))
+if test x"$enable_mmap" = x"yes"; then
+    AC_MSG_RESULT(yes)
+    AC_DEFINE(ENABLE_MMAP, 1, [Define if you want mmap.  By default this is undefined.])
+else
+    AC_MSG_RESULT(no)
+fi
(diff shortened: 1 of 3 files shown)


Alexponomarev7:
@Wayne could you please share some of your thoughts on this? If this has a chance of life, I will add tests/improve the patch, etc.

WayneD:
The comments in fileio.c indicate that mmap() was avoided because it can lead to a SIGBUS when a file gets truncated while it is being copied.  The current fileio code treats the read errors that occur due to a truncated file as if all the imaginary bytes are zeros and continues.  To be able to do that with mmap() would require some kind of signal recovery code, such as using sigsetjmp() and siglongjmp().  It would also require that the buffer-accessing operations themselves get protected, so the callers of map_ptr() would either need to be given a copy of the bytes they requested (ensuring that they can't possibly get a SIGBUS once map_ptr() returns) or the signal-protection code would need to be extended into each caller's code with some kind of recovery logic (which looks to be pretty hard to do for things like `write_buf(f, map_ptr(...))`).  Of course, the final possibility is that the mmap() version of rsync just accepts that the entire transfer fails if something truncates a file while it is being read.

Here's test program that generates a SIGBUS attempting to read all the bytes of a truncated file:
[mmap-fail.c.txt](https://github.com/WayneD/rsync/files/9795964/mmap-fail.c.txt)

PR #393, opened 2022-10-23
Update and rename .github/workflows/build.yml to .github/سير العمل/bu…

basl196:
…ild.yml


basl196:
https://github.com/WayneD/rsync/pull/393#issue-1419780699

PR #426, opened 2023-01-01
configure.ac: use pkg-config to retrieve openssl dependencies

ffontaine:
Use pkg-config to retrieve openssl dependencies such as `-latomic` and avoids the following build failure when building statically on architectures such as sparc:

```
/home/autobuild/autobuild/instance-1/output-1/host/lib/gcc/sparc-buildroot-linux-uclibc/10.4.0/../../../../sparc-buildroot-linux-uclibc/bin/ld: /home/autobuild/autobuild/instance-1/output-1/host/sparc-buildroot-linux-uclibc/sysroot/usr/lib/libcrypto.a(threads_pthread.o): in function `CRYPTO_atomic_add': threads_pthread.c:(.text+0x208): undefined reference to `__atomic_is_lock_free'
```

Fixes:
 - http://autobuild.buildroot.org/results/49abbaa1eab94b248bff434b40728065d687e278

Signed-off-by: Fabrice Fontaine <fontaine.fabrice@gmail.com>

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -432,11 +432,15 @@ AH_TEMPLATE([USE_OPENSSL],
 if test x"$enable_openssl" != x"no"; then
     if test x"$ac_cv_header_openssl_md4_h" = x"yes" && test x"$ac_cv_header_openssl_md5_h" = x"yes"; then
       AC_MSG_RESULT(yes)
-      AC_SEARCH_LIBS(EVP_MD_CTX_copy, crypto,
+      PKG_CHECK_MODULES(LIBCRYPTO, libcrypto,
           [AC_DEFINE(USE_OPENSSL)
-	   enable_openssl=yes],
-          [err_msg="$err_msg$nl- Failed to find EVP_MD_CTX_copy function in openssl crypto lib.";
-	   no_lib="$no_lib openssl"])
+           enable_openssl=yes
+           LIBS="$LIBS $LIBCRYPTO_LIBS"],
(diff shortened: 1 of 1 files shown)


ffontaine:
Do you have some comments on this PR?

tridge (changes requested):
with autoconf 2.71 I get this: ```
checking whether to enable use of openssl crypto library... ./configure: line 7719: syntax error near unexpected token `LIBCRYPTO,'
./configure: line 7719: `      PKG_CHECK_MODULES(LIBCRYPTO, libcrypto,'
```
maybe bitrotted?

ffontaine:
Thanks for your feedback.

Patch has been applied in buildroot since February (https://git.buildroot.net/buildroot/commit/?id=e889a1c9e983753dd0fa5062d3b9475a8cba6072) and no build failures are raised by autobuilders since that time.

It should be noted that `configure` must be rebuilt (thanks to `autoreconf `) after applying this patch. which updates `configure.ac`. 
Could you also try to add `PKG_PROG_PKG_CONFIG `before  `PKG_CHECK_MODULES`? This macro will check if pkg-config is available on your system.

tridge:
@ffontaine it was fine after "make reconfigure". Note sure why autoconf didn't generate a good configure script for me yesterday. It's been many years since I worked with autoconf - about 20 years in fact, when I handed over to Wayne :-)

tridge (approved):

tridge:
@WayneD I'm happy with this

vilkov:
Hey @tridge, @WayneD can we have this merged, please?

PR #459, opened 2023-03-21
improve human readable output

eworm-de:
Calculate in a loop, use dynamic precision.


eworm-de:
As there is a new pre-release... Can I have a quick feedback and state on this?

WayneD:
The current release only has bugfixes in the main code (it's only the adjunct scripts & manpages that got enhancements).

I'm very cautious in making output-affecting changing, but will be giving this a closer look.

eworm-de:
I can understand you are cautious about making output-affecting changing, guess this is due to breaking scripts that expect specific output?
My changes effect _human_ readable output only, though. 😉 I hope this will not break anything.

tridge:
@eworm-de could you add a description of the before and after of what this change does? Needs a bit more context than "improve" :-)

eworm-de:
The commit messages have the details. Are you missing something there or just here in the pull request?

eworm-de:
This is the output of `git log master..`:
```
commit 75ea671e822b347cd4a6656d9913eb9fa89ff600 (HEAD -> human)
Author: Christian Hesse <mail@eworm.de>
Date:   Thu Mar 23 10:33:33 2023 +0100

    human-readable: add an "i" to unit to indicate binary (1024) base

commit 4f215ad2cf7b0dff74db74202412793a9947cadb
Author: Christian Hesse <mail@eworm.de>
Date:   Wed Mar 22 11:07:15 2023 +0100

    human-readable: also handle num < mult with the same code
    
    ... just make sure no precision is added.

commit d4784655035dcfdf02efb30164b181f47546713b
Author: Christian Hesse <mail@eworm.de>
Date:   Tue Mar 21 10:17:09 2023 +0100

    human-readable: use dynamic precision length
    
    Let's lower precision for huge numbers.
    
    The output used to be:
    
      3.45M -> 46.73M -> 523.11M -> 1.24G -> ...
    
    With this change the code always gives the three most reasobable digits:
    
      3.45M -> 46.7M -> 523M -> 1.24G -> ...

commit 53e4e68e9d5d195679db74e178c4d9e915094236
Author: Christian Hesse <mail@eworm.de>
Date:   Tue Mar 21 10:13:47 2023 +0100

    human-readable: calculate numbers in a loop
    
    This drops a lot of `else if` blocks and extends units
    by "E", "Z" & "Y".
```

tridge:
@eworm-de ok, thanks, I normally expect it in the PR message
@WayneD are we concerned about users having parses these messages? Also, should we have a unit test for this?

eworm-de:
Just added another commit:
```
commit 7f26b1d2dd34ec39a566d0ba42a43bf22d6d5c46 (HEAD -> human, origin/human)
Author: Christian Hesse <mail@eworm.de>
Date:   Tue Apr 9 09:56:39 2024 +0200

    human-readable: also use it to format rate
    
    Let's also simplify the code for rate in progress, and benefit from
    same functionality.
```

eworm-de:
> are we concerned about users having parses these messages?

I still stand by my point: This is about *human* readable output. We should not be worried too much about compatibility with parsers.

eworm-de:
Any news on this? Does it help if I split this across several pull requests?

eworm-de:
Now that we have a fresh release... Any chance for this?

eworm-de:
This is now three years old and did not see any progress. Does it help if I split into smaller pull requests?

progress.c
```diff
@@ -93,41 +92,31 @@ static void rprint_progress(OFF_T ofs, OFF_T size, struct timeval *now, int is_l
 		/* Compute stats based on the starting info. */
 		if (!ph_start.time.tv_sec || !(diff = msdiff(&ph_start.time, now)))
 			diff = 1;
-		rate = (double) (ofs - ph_start.ofs) * 1000.0 / diff / 1024.0;
+		rate = (ofs - ph_start.ofs) * 1000 / diff;
 		/* Switch to total time taken for our last update. */
-		remain = (double) diff / 1000.0;
+		remain = diff;
```
steadytao: This keeps `diff` in milliseconds but the formatter below still treats `remain` as seconds. The old code divided by `1000.0` here. With this change, a final progress line for an 8 second transfer would display about 8000 seconds of elapsed time.

lib/compat.c
```diff
@@ -173,36 +173,44 @@ char *do_big_num(int64 num, int human_flag, const char *fract)
...
+		if (powi == 1)
+			precision = 0;
+
+		for (; precision > 0; precision--) {
+			powj *= 10;
+			if (abs_num / powi < powj)
+				break;
+		}
+
+		snprintf(bufs[n], sizeof bufs[0], "%.*f%c%s", precision,
```
steadytao: For values below the first unit threshold, `*units` is the leading space in `" KMGTPE"` so this returns strings such as `"999 "` instead of `"999"`. Since `human_num()` is shared outside progress output this leaks a trailing blank into list/stat/log formatting rather than just producing the intended `999B` progress display.

lib/compat.c
```diff
@@ -173,36 +173,44 @@ char *do_big_num(int64 num, int human_flag, const char *fract)
...
+			(double) num / powi, *units, num > mult && mult == 1024 ? "i" : "");
```
steadytao: This should probably use `abs_num >= mult` rather than `num > mult`. As written, level-3 output omits the `i` suffix for exact-threshold values such as `1024` and for negative values such as `-2048` even though those are still binary-unit outputs.

steadytao (changes requested):
Thanks for sticking with this. I like the direction of making the formatter less hand-coded but I would request changes on the current patch.

Alongside my inline comments, the manpage progress examples are now stale and the current tests dont appear to cover the exact human-readable/progress output being changed here. Given this changes user-visible output, I think it needs focused formatter/progress coverage before merge alongside the bug fixes.

progress.c
eworm-de: Hmm, no idea why I dropped that... Added in again.

lib/compat.c
eworm-de: Ah, right... This did not matter a lot when I added this. The trailing space was there, but not visible in output. That changed with later commits. Fixed now, handling this as string, not printing just a character.

lib/compat.c
eworm-de: Now that the unit is a terminated string... Let's just check for `*units`. 😊️

lib/compat.c
eworm-de: Wait... That does not work. Truncating the string with `units[1] = '\0'` does not bring the expected result... 🤨️

eworm-de: Ok, fixed now.

lib/compat.c
```diff
@@ -173,36 +173,46 @@ char *do_big_num(int64 num, int human_flag, const char *fract)
...
+		if (powi == 1)
+			precision = 0;
+
+		for (; precision > 0; precision--) {
+			powj *= 10;
+			if (abs_num / powi < powj)
+				break;
+		}
+
+		snprintf(bufs[n], sizeof bufs[0], "%.*f%s%s", precision,
```
steadytao: This still changes the below-threshold `do_big_dnum()` behaviour for level-3 output. `do_big_dnum()` passes `fract` here to preserve decimals when the value is not actually unit-scaled. With `human_flag == 3`, values such as `1000.12` and `1023.12` are below the 1024 threshold but this branch now formats them with `precision = 0` and ignores `fract` producing `1000` / `1023` instead of preserving the decimal part.

progress.c
```diff
@@ -93,41 +92,31 @@ static void rprint_progress(OFF_T ofs, OFF_T size, struct timeval *now, int is_l
...
+		remain = diff / 1000;
 	} else {
 		strlcpy(eol, "  ", sizeof eol);
 		/* Compute stats based on recent progress. */
 		if (!(diff = msdiff(&ph_list[oldest_hpos].time, now)))
 			diff = 1;
-		rate = (double) (ofs - ph_list[oldest_hpos].ofs) * 1000.0 / diff / 1024.0;
-		remain = rate ? (double) (size - ofs) / rate / 1000.0 : 0.0;
+		rate = (ofs - ph_list[oldest_hpos].ofs) * 1000 / diff;
+		remain = rate ? (size - ofs) / rate : 0;
```
steadytao: This now does signed integer multiplication before division. For very large files or progress2 totals, `(ofs - ph_start.ofs) * 1000` can overflow before `diff` reduces it. The old calculation used `double` and avoided that intermediate overflow. If the final display wants `human_num(rate)` the rate can still be converted after computing it in a wider/safer way.

steadytao (changes requested):
Quick overall thing as well as the inlines. The manpage progress examples are still stale and the tests still only check very broad output properties. Since this PR intentionally changes user-visible formatting, I think it needs very focused tests for the formatter and at least one exact progress/final-stats shape before merge.

eworm-de:
> The manpage progress examples are still stale 

Ups, I was not aware that the dynamic precision never made it into  the man page. Updated... Is that sufficient?

> and the tests still only check very broad output properties. Since this PR intentionally changes user-visible formatting, I think it needs very focused tests for the formatter and at least one exact progress/final-stats shape before merge.

Agreed. But the test suite is (mostly?) python, and I am not very familiar with that. Any help appreciated...

lib/compat.c
eworm-de: Hmm, I do not touch any code that uses `fract`. How is this relevant? What code path and what data causes issues here, and is that different from before?

progress.c
eworm-de: Oh, `OFF_T` is just 32 bits? I did not expect that. Restoring the cast... Is it ok now?

lib/compat.c
steadytao: Thanks. The path is final stats, not progress.

`bytes_per_sec_human_dnum()` calls `human_dnum(..., 2)` which goes through `do_big_dnum()` and then calls `do_big_num((int64)dnum, human_flag, fract)`. For level-3 human output, values like `1000.12` or `1023.12` are below the 1024 binary-unit threshold.

Previously those values fell through to the raw formatter and preserved `fract`. With the current branch, the new formatter handles them with `powi == 1`, precision `0` and no `fract` so the fractional part is lost. I think the smallest fix is to let `powi == 1` fall through to the existing raw formatter.

progress.c
steadytao: Yes, restoring the floating-point calculation addresses the overflow concern. I would keep `rate` and `remain` as `double` through the calculation and only cast `rate` at the final `human_num()` call so the elapsed/remaining time maths keeps the previous behaviour too.

steadytao (approved):
LGTM. Awaiting @tridge review.

eworm-de:
Oh, you pushed commit into my pull request. Gladly I noticed before I force-pushed my rebase. 😄️

Some of your changes are more cleanups for my commits. Is it ok for you if I clean up the history a bit?

Also some of the code is heavily influenced by you. Are you fine if I add `Co-authored-by` for you into my commits?

steadytao:
> Some of your changes are more cleanups for my commits. Is it ok for you if I clean up the history a bit?

Feel happy to. You can even make changes if you notice anything. It is your branch 😁

> Also some of the code is heavily influenced by you. Are you fine if I add `Co-authored-by` for you into my commits?

I wouldn't mind either way but thank you for the offer, of course you can!

eworm-de:
Shuffled a bit... The result should be the same. Again, thanks a lot!

steadytao:
Thank you for your work!

progress.c
```diff
@@ -126,8 +115,8 @@ static void rprint_progress(OFF_T ofs, OFF_T size, struct timeval *now, int is_l
 
 	output_needs_newline = 0;
 	pct = ofs == size ? 100 : (int) (100.0 * ofs / size);
-	rprintf(FCLIENT, "\r%15s %3d%% %7.2f%s %s%s",
-		human_num(ofs), pct, rate, units, rembuf, eol);
+	rprintf(FCLIENT, "\r%15sB %3d%% %7sB/s %s%s",
+		human_num(ofs), pct, human_num((int64)rate), rembuf, eol);
```
steadytao, line 119: I missed the range check when I suggested casting at the final formatting call. A sufficiently large byte span with a short elapsed time produces a rate above INT64_MAX so this double-to-int64 conversion is undefined. Please check the range before casting or retain floating-point formatting for out-of-range rates and add a boundary regression.

steadytao (changes requested):

steadytao:
Ah also, just had a thought on the co-authored. You can leave commits authored by yourself but to ensure the tree is verified rather then unverified, you could remove me and I can add co-authored on squash if that suits?

progress.c
eworm-de: Updated... Does that account for a sufficient check?

eworm-de:
> Ah also, just had a thought on the co-authored. You can leave commits authored by yourself but to ensure the tree is verified rather then unverified, you could remove me and I can add co-authored on squash if that suits?

Hmm, I can not follow... The co-authored-line is just text in the commit message. Why remove that?

Or did I get you wrong? Just enabled gpg-signing for the repository, commits are shown as *partially verified* now at least. You would have to force-rebase to make yourself the committer, that would also add your signatures.

I put a lot of effort in creating a meaningful history. I would prefer not to squash this... Why would you?

PR #465, opened 2023-04-05
hint that a proxy can handle plain and ssl stream at the same time

eworm-de:


rsyncd.conf.5.md
```diff
@@ -1188,6 +1188,9 @@ An example nginx proxy setup is as follows:
 > }
 > ```
 
+If rsyncd shoud be accessible encrypt and unencrypted at the same time make
```
tridge, line 1191: minor typos, "should", "encrypted"

tridge (changes requested):

tridge (approved):

PR #477, opened 2023-05-03
Introduce PTR_SUB

AreaZR:
This is more intuitive than adding a negative number.


tridge (approved):

tridge:
@WayneD looks good, please check

PR #501, opened 2023-07-02
Fix runtime AVX2 detection for checksum

Icenowy:
The way GCC implements function-multiversioning is to generate a dispatcher function, thus it could not be used when a function is not built by GCC (e.g. a direct .S file, in the situation of AVX2 checksum).

Use function-multiversioning to declare a wrapper that will calls the appropriate checksum generatator function when AVX2 is present and do nothing when not present.

### DIFF
--- a/simd-checksum-x86_64.cpp
+++ b/simd-checksum-x86_64.cpp
@@ -85,9 +85,7 @@ typedef long long __m256i_u __attribute__((__vector_size__(32), __may_alias__, _
 #define SSE2_HADDS_EPI16(a, b) _mm_adds_epi16(SSE2_INTERLEAVE_EVEN_EPI16(a, b), SSE2_INTERLEAVE_ODD_EPI16(a, b))
 #define SSE2_MADDUBS_EPI16(a, b) _mm_adds_epi16(SSE2_MULU_EVEN_EPI8(a, b), SSE2_MULU_ODD_EPI8(a, b))
 
-#ifndef USE_ROLL_ASM
-__attribute__ ((target("default"))) MVSTATIC int32 get_checksum1_avx2_64(schar* buf, int32 len, int32 i, uint32* ps1, uint32* ps2) { return i; }
-#endif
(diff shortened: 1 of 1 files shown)


hhoffstaette:
I know it's been a while but I just found that this breaks simdtest in 3.4.3 with an illegal instruction on my SandyBridge machine (where I build with -march=sandybridge, which has AVX but no AVX2). I first suspected gcc-16 but it also happens with clang-22; removing this patch makes the test pass. Everything is fine on my Zen2 laptop, which has AVX2.

hhoffstaette:
For some reason simdtest ends up in `get_checksum1_avx2_64`:
```
$rm simdtest && make simdtest && ./simdtest
if test x"simd-checksum-x86_64.o" != x; then \
    x86_64-pc-linux-gnu-g++ -I. -pipe -O2 -march=sandybridge -fno-semantic-interposition -fno-exceptions -fno-rtti  -Wl,-O1 -Wl,--as-needed -Wl,-z,pack-relative-relocs -DTEST_SIMD_CHECKSUM1 \
        -o simdtest ./simd-checksum-x86_64.cpp  -lacl -lpopt -lzstd -lxxhash -lcrypto -lz ; \
else \
    touch simdtest; \
fi
zsh: illegal hardware instruction  ./simdtest

$gdb simdtest 
Reading symbols from simdtest...
(No debugging symbols found in simdtest)
(gdb) r
Starting program: /tmp/portage/net-misc/rsync-3.4.3-r1/work/rsync-3.4.3/simdtest 
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib64/libthread_db.so.1".

Program received signal SIGILL, Illegal instruction.
0x0000555555556df0 in get_checksum1_avx2_64(signed char*, int, int, unsigned int*, unsigned int*) [clone .part.0] ()
```

Probably a bad interaction with https://github.com/RsyncProject/rsync/commit/aa142f08ef31d3ffa8d6b3b8af16d00324a98c1b

tridge:
@Icenowy ahh, apologies, I just fixed an overlapping issue with AVX2 reported by someone testing the 3.5 security test releases, can you rebase and see if there is anything left that needs fixing?

PR #521, opened 2023-09-04
Fix misc. warnings found by clang-16

hhoffstaette:
These warnings may seem benign but will soon turn into compilation errors, starting with clang-17 and gcc-14. I tried hard to find a use for the duplicated (and incorrect) lseek prototypes, but could not find any.
The alignment issue was also flagged by clang. I'm not a SIMD expert but explicit 1-byte aligment for a vector seems wrong.

With these fixes rsync itself is now 100% warning-free :)


tridge (approved):

tridge:
looks good, assigning to @WayneD for merge

PR #536, opened 2023-10-09
fix a possible memory leak if execvp fails in execCommand

DuratarskeyK:
argv isn't freed, if execvp fails at the end of the function, creating a memory leak.


tridge (changes requested):
popt is based on an upstream version, can you point at this change or equivalent upstream?
otherwise we would be calling exit anyway, so memory leak seems unimportant

PR #539, opened 2023-10-12
Make size related variables and outputs size_t

AZero13:
It makes sense for these values that are expected to be unsigned to be treated as such, especially on 64 bit platforms where we can avoid the cost of truncation by only doing it when needed, as opposed to every time strlen returns a value


acls.c
```diff
@@ -758,14 +761,15 @@ static int recv_rsync_acl(int f, item_list *racl_list, SMB_ACL_TYPE_T type, mode
...
 		duo_item->racl.other_obj = recv_acl_access(f, NULL);
-	if (flags & XMIT_NAME_LIST)
-		computed_mask_bits |= recv_ida_entries(f, &duo_item->racl.names);
 
 #ifdef HAVE_OSX_ACLS
 	/* If we received a superfluous mask, throw it away. */
 	duo_item->racl.mask_obj = NO_ENTRY;
 	(void)mode;
 #else
+	if (flags & XMIT_NAME_LIST)
```
tridge, line 770: this change looks like it will change the wire protocol if HAVE_OSX_ACLS is defined

tridge (changes requested):
this PR seems to be doing rather more than what the commit msg describes?

PR #540, opened 2023-10-12
Optimize check_batch_flags

AtariDreams:


tridge:
i'd be very surprised if this is actually faster, and it is a function where speed doesn't matter, I don't think this is worth the (admittedly small) chance of unintended consequences, closing

PR #554, opened 2023-12-15
Add support/rsync3: like 'scp -3' but with rsync full power

zerodeux:
Hello,

I'm often asked how to move files between disjoint network zones, usually through one's own computer or a specialized one like a 'bastion', and came to write a simple wrapper which let one use rsync as usual, but allowing remote-to-remote communications.

This uses the 'scp -3' model where trafic between the 2 remotes flow through the one which started the command, there is no direct sourece-to-destination trafic (since in segregated networks it's not possible). From my experience it's a more and more common scenario, and it's even hard to find good tips/hacks on Stackoverflow and such. This script makes it easy and handy to solve the problem.

It obviously only works with the SSH transport, and it's clearly stated in the script's help text.

### DIFF
--- a/support/rsync3
+++ b/support/rsync3
@@ -1,0 +1,77 @@
+#!/bin/bash
+
+# Remote-to-remote invocation of rsync, with the help of SSH's -R.
+# Use like 'rsync', but you're allowed to specify a 'remote:' on both ends.
+#
+# Copyright (C) 2023  Bearstech
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+# Pick a listening port on the source target, let's hope it's available.
+# (only binds on 127.0.0.1, should not have security/firewalling issues)
+port=45997
+
+# Turn $@ into a list because it's easier to pick the last elements this way
+# Also intercept our own flag(s).
+opt=()
+opt_nocheck=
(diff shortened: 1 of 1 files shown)


SIGSEGV111:
Wouldn't this be easier with ProxyJump from SSH? (check `man ssh_config`).

zerodeux:
I'm not sure how SSH's ProxyJump would help to solve this problem in a single command. The problem is not about "bouncing" an SSH access, but establishing a tunnel between 2 isolated endpoints, through a 3rd party.

SIGSEGV111:
Lets say you want to rsync between your localhost and a host called "bastion", which you can only reach through a host called "gatekeeper".

You would create a entry in `ssh_config` like this:

```
Host bastion
    ProxyJump gatekeeper
```

now you can ssh directly to bastion like so:
`ssh bastion`
from your local machine and let ssh figure out all the "complicated" forwardings.

Since rsync uses ssh as transport, you can now also use rsync on bastion as well:
`rsync -vra /local/path/ bastion:/some/path/`

Naturally you can configure all (?) config-entries in `ssh_config` also via command-line:
`ssh -oProxyJump=gatekeeper bastion`

using this in rsync would yield:
`rsync --rsh='ssh -oProxyJump=gatekeeper'  -vra /local/path/ bastion:/some/path/`

Hope this help!

PS: I typed all of the above from memory with only the briefest look at the man page. The syntax might be a off, but the concept should work.

zerodeux:
Thanks for the clear explanations.

I do use ProxyJumps in my `.ssh/config`, but I don't think `rsync3` solves the same problem.

Rsync3 is meant to solve remote to remote transfers where both ends are not able to connect to each other, which is non-trivial to setup (you need at least a reverse-tunnel, bridging the two isolated ends).

In more details, rsync3 will connect to host A (:22/ssh) :
* running the rsync client part
* starting a reverse tunnel to B (:22/ssh), which means that your workstation will actually do the relaying between A and B for this tunnel

In turn the rsync client part on A will connect to B through the tunnel, using the usual `-e 'ssh -p port host'` trick.

I don't think you can automate this with config files and it's pretty tricky to correctly invocate, although in the end it's ineed 1 proper invocation of SSH and 1 proper invocation of rsync. I've seen several sysadmins struggle with this and I wanted console-wary developers to be able to use this. In my experience, `rsync3` did that trick properly.

PR #558, opened 2024-01-02
configure.ac: fix failing IPv6 check due to missing return type

bobrik:
Fixing this warning escalated to an error, resuting in no IPv6 support:

```
configure.sh:7679: checking whether to enable ipv6
configure.sh:7718: clang -o conftest -g -O2 -DHAVE_CONFIG_H -Wall -W   conftest.c  >&5
conftest.c:73:1: error: type specifier missing, defaults to 'int'; ISO C99 and later do not support implicit int [-Wimplicit-int]
main()
^
int
1 error generated.
configure.sh:7718: $? = 1
configure.sh: program exited with status 1
```

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -392,7 +392,7 @@ AS_HELP_STRING([--disable-ipv6],[disable to omit ipv6 support]),
 #include <stdlib.h>
 #include <sys/types.h>
 #include <sys/socket.h>
-main()
+int main()
 {
    if (socket(AF_INET6, SOCK_STREAM, 0) < 0)
      exit(1);


bobrik:
It was a warning on `clang-15`, but it is an error on `clang-16`:

* https://github.com/NixOS/nixpkgs/pull/278184

bobrik:
A similar thing happened to curl: https://github.com/curl/curl/pull/12607

bobrik:
I'm happy to pull the curl's version verbatim if you think that's better.

tridge (approved):

tridge:
@bobrik looks good, thanks!

PR #586, opened 2024-03-29
Allow basic connectivity check via rrsync

cjwatson:
rsbackup (https://github.com/ewxrjk/rsbackup) uses `ssh <host> true` to check that the host in question is reachable.  I like to configure my backed-up hosts to force the backup system to go via `rrsync`, but I always have to add a local tweak to allow `SSH_ORIGINAL_COMMAND=true` to work.  I think this would be safe enough to include in rrsync.


tridge (approved):

tridge:
@WayneD looks good to me, can you check too?

PR #595, opened 2024-04-08
ignore-non-existing-directory: add variant of ignore-non-existing

robbat2:
Add new option --ignore-non-existing-directory, that is a variant of --ignore-non-existing, but applies ONLY to directories.

This was previously proposed in bug #8366, but I independently had a use case for it in the Gentoo infrastructure.

Reference: https://bugzilla.samba.org/show_bug.cgi?id=8366
Reference: http://superuser.com/questions/316561/rsync-synchronizing-files-only-without-creating-folders-on-destination
Reference: https://github.com/RsyncProject/rsync-patches/issues/8
Reference: https://lists.samba.org/archive/rsync/2015-November/030455.html

### DIFF
--- a/generator.c
+++ b/generator.c
@@ -60,6 +60,7 @@ extern int update_only;
 extern int human_readable;
 extern int ignore_existing;
 extern int ignore_non_existing;
+extern int ignore_non_existing_dirs;
 extern int want_xattr_optim;
 extern int modify_window;
 extern int inplace;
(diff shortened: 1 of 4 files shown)


ToastMaster1014 (changes requested):
Change 6

robbat2:
@tridge another PR for merge please?

ToastMaster1014:
Ignore bitch I will redirect you again

ToastMaster1014:
> @tridge another PR for merge please?

No sorry can't do that.

PR #601, opened 2024-04-19
Apple silicon linker path details

vaguely-tagged:
I added some notes for under macos for the install.md to make it a bit more clear how to link libraries for m series macs due to the new non standard install location of homebrew packages.


charmitro:
As a macOS user, this note is correct.

tridge (approved):

PR #608, opened 2024-04-29
add options to don't use bunbled libs

dartmol203:
I was maintaining the Debian package of rsync and i had to add this changes as a patch and I think it would be a good addition to the project


tridge:
@dartmol203 the problem with not using the bundled zlib is rsync by default uses a modification to zlib that allows for better compression ratios by compressing data that is not sent over the link so that the compression history is better setup for the data.
If you use the non-bundled zlib this this will break compatibility.
Unfortunately I have noticed that distros like debian have gone ahead and made this change anyway, which leaves us with the messy situation of incompatible rsync versions.

WayneD:
I'm not understanding why the existing configure options aren't working for you. The change you made just seems to prevent the zlib/dummy file & popt/dummy files from being created. Those dummy files won't affect what libraries are configured into rsync. The existing `--without-included-popt` and `--without-included-zlib` options already work fine. Are you just trying to neaten up the build dir?

WayneD:
Some comments on the old zlib issue for Tridge's sake: the modern rsync avoids the implied compression data when using zlib compression in order to be more compatible with a wide range of zlib implementations. I did get support for the rsync kluge into the official zlib, but given that rsync compression seems to be somewhat fragile, the default is to avoid the old-style implied compress data (since it seems like the official library might not be 100% compatible with the included popt in some scenarios). In the most modern rsyncs, a transfer negotiates a compression algorithm from the list of "zstd lz4 zlibx zlib" where zlibx is the newer style of zlib compression. This means that the newest rsync supports both kinds of zlib compression, even with an external zlib, but defaults to the newer style of zlib.

samueloph:
Hello everyone,

@tridge 
> If you use the non-bundled zlib this this will break compatibility.
> Unfortunately I have noticed that distros like debian have gone ahead and made this change anyway, which leaves us with the messy situation of incompatible rsync versions.

It has been a few years since I've switched Debian's rsync to dynamically link to plain zlib, so my memory is not perfect, but I recall that I did extensive testing to ensure there was compatibility by making transfers between differently built rsyncs.

I also have a brief memory of Wayne writing somewhere that it should be fine to make the switch. It was probably what he mentioned in a later comment on this thread: https://github.com/RsyncProject/rsync/pull/608#issuecomment-2138312832

All of this being said, I'm happy to take feedback on this and consider reverting the change if you think that's the right approach (again, considering Wayne's comment, sounds like we're fine).

@WayneD 
> I'm not understanding why the existing configure options aren't working for you.
> Are you just trying to neaten up the build dir?

We are looking for support for the following scenario:
1) Get the rsync tarball, or clones the repo.
2) Remove the zlib and popt folders from rsync's sources
3) Build rsync with the flags `with_included_popt=no` and `with_included_zlib=no`

The PR adds support for this, without any downsides attached.

Now, the reason for explicitly removing those folders is two-fold:
1) It makes impossible for one to mistakenly compile with the bundled copy.
2) It tidies up the distributed source code, less files, less licenses/copyright holders involved.

WayneD:
As I pointed out, the existing configure options work just fine without any patches. The only thing this diff seems to do is to not create "empty" dirs (that contain just a "dummy" file) in the build dir, which might be nice but isn't necessary. Those empty dirs don't contain any built files and are ignored.

PR #623, opened 2024-06-23
Most open-source software use the en_US speling [sic]

dotysan:
- apologies to the folks in the UK

### DIFF
--- a/usage.c
+++ b/usage.c
@@ -313,7 +313,7 @@ void print_rsync_version(enum logcode f)
 	rprintf(f,"\n");
 	rprintf(f,"rsync comes with ABSOLUTELY NO WARRANTY.  This is free software, and you\n");
 	rprintf(f,"are welcome to redistribute it under certain conditions.  See the GNU\n");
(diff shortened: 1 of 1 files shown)


michaelmess-de:
https://www.gnu.org/licenses/gpl-3.0.html
Yes the correct spelling when referencing that document is License.

But it has nothing to do with that most software may use en_US spelling, relevant is only the name of the license document that we are referring to, which is called "GNU General Public License".

Thus I would change the title of the pull-request to something else, more meaningful, e.g.: "Correct spelling in reference to GNU General Public License".

By the way, there are 3 versions of the GNU General Public License.
 - https://www.gnu.org/licenses/old-licenses/gpl-1.0.html
 - https://www.gnu.org/licenses/old-licenses/gpl-2.0.html
 - https://www.gnu.org/licenses/gpl-3.0.html
It is then still not clear which one of that three versions is actually meant here.
But in file headers, version 3 or any later one is referenced to (with correct spelling), maybe we should add that information here, too.

PR #639, opened 2024-09-05
Return from #list command with 0

tacerus:
The "#list" command should not be treated as a failure when it is both a legitimate request by the client, and correctly answered by the server. It is commonly used for assessing whether a rsync endpoint is healthy, having it return with a non-zero exit code causes misleading error reports, and, in case of socket activation, failed service instances on the server.


tridge:
what testing has been done? Have you tested both inetd operation and running as a daemon?

tacerus:
I tested running through a socket activated systemd service.

PR #687, opened 2025-01-03
Fix generated man pages to make searching of long options compatible with more browsers

### LINKED ISSUE #686: Certain hyphenated options on rsync manual page are unsearchable
Observed with 
```
$ rsync --version
rsync  version 3.3.0  protocol version 31
```
on 64-bit Cygwin.

Steps to reproduce:
- Run `man rsync`.
- Search for `no-compress`. You find no results.
- Search for `implies --no`. You find:
```
This  option can be used to override the automatic negotiation of
the compression algorithm that occurs when  --compress  is  used.
The  option implies --compress unless "none" was specified, which
instead implies --no‐compress.
```
I sure see `--no-compress` on the bottom. This is just one example. There are a couple other unsearchable hyphens in the manual page as well.

(Conjectured) Cause:
The `nroff` format that is used to write man pages has the following peculiarity:
- The raw ASCII hyphen `-` in the nroff source is rendered as U+2010 HYPHEN.
- The escaped ASCII hyphen `\-` in the nroff source is rendered as an ASCII hyphen.

Inspecting the source of `rsync.1`, we see that the relevant hyphens were not escaped:
```
This option can be used to override the automatic negotiation of the
compression algorithm that occurs when \fB\-\-compress\fP is used.  The
option implies \fB\-\-compress\fP unless "none" was specified, which
instead implies \fB\-\-no-compress\fP.
```

purplearmadillo77:
I think this fixes #686. The `NON_SPACE_SINGLE_DASH_RE` regular expression was matching non-word followed by dash (`(^|\W)-`) rather than what the variable name suggests.


WayneD:
This transforms all hyphenated words, not just long options. So, no hyphenated words would be breakable.

I'll also note that the words are searchable in Chrome, since chrome will cross-match a dash and that hyphen. It's sad that not all browsers do this.

WayneD:
I tweaked the script with a new regex that should handle this issue in long options while leaving hyphenated words alone.

WayneD:
I committed my change after making a regex improvement.

Issue #637, opened 2024-08-29
PC via ssh sync to Android, report: `rsync: symlink ...  failed: Permission denied (13)`


xiaguangbo:
Android ssh app: SimpleSSHD

```
$ rsync -av --delete -e 'ssh -p 2222' ~/xiaguangbo 192.168.1.118:/sdcard
...
rsync: symlink "/sdcard/xiaguangbo/data/esp/espressif/tools/riscv32-esp-elf/esp-13.2.0_20240530/riscv32-esp-elf/lib/gcc/riscv32-esp-elf/13.2.0/plugin/libcp1plugin.so" -> "libcp1plugin.so.0.0.0" failed: Permission denied (13)
...
rsync: symlink "/sdcard/xiaguangbo/project/tempctrl/program_esp32c2/project/target/riscv32imc-esp-espidf/release/build/esp-idf-sys-d9828311749dae4a/out/build/esp-idf/mbedtls/mbedtls/library/ssl_debug_helpers_generated.c" -> "/home/xiaguangbo/桌面/data/esp/espressif/esp-idf/v5.3/components/mbedtls/mbedtls/library/ssl_debug_helpers_generated.c" failed: Permission denied (13)
```

realsimix:
What is the filesystem on the sdcard? Does it even support symlinks?

xiaguangbo:
`-a` == `-rlptgoD`
use `L`, no `l`. use real file, no symlink
`rsync -rLptgoDvziPh --delete -e 'ssh -p 2222' ~/xiaguangbo 192.168.1.126:/sdcard`

PR #693, opened 2025-01-13
un-initialization of a static char * (fix of issue #637)

lighterCoder:
to ininitialize buf1 in case it is not initiliazed


checksum.c
```diff
@@ -361,7 +361,7 @@ void get_checksum2(char *buf, int32 len, char *sum)
 	  case CSUM_MD4_ARCHAIC: {
 		md_context m;
 		int32 i;
-		static char *buf1;
+		static char *buf1 = NULL;
```
tridge, line 364: static variables are always initialised to NULL, so this is not needed

tridge:
not needed for static vars

PR #705, opened 2025-01-15
Fix FLAG_GOT_DIR_FLIST collission with FLAG_HLINKED

ncopa:
fixes commit 688f5c379a43 (Refuse a duplicate dirlist.)

Fixes: https://github.com/RsyncProject/rsync/issues/702
Fixes: https://github.com/RsyncProject/rsync/issues/697


tridge (approved):
thanks for looking at this so quickly!
I don't suppose you want to to become an rsync maintainer?

ncopa:
> thanks for looking at this so quickly! I don't suppose you want to to become an rsync maintainer?

Not really. But as downstream maintainer I don't mind help when necessary.

PR #707, opened 2025-01-15
Fix build on ancient glibc without openat(AT_FDCWD

ncopa:
Fixes: https://github.com/RsyncProject/rsync/issues/701

### DIFF
--- a/syscall.c
+++ b/syscall.c
@@ -734,7 +734,7 @@ int secure_relative_open(const char *basedir, const char *relpath, int flags, mo
 		return -1;
 	}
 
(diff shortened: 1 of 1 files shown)


solardiz (approved):

solardiz:
Separately, maybe we could add something like:

```c
#ifdef __GNUC__
#warning "System features required for security missing, CVE-... CVE-... remain unfixed"
#endif
```

At least this would be triggered if `AT_FDCWD` ever stops being a macro (hopefully, never).

tridge (approved):

PR #719, opened 2025-01-20
Add option --time-only

steinhh:
We maintain an archive with Solar observations (46 million and counting), with the files originating at another site (Japan). Naturally, we use rsync to fetch the files. Some observations are stored uncompressed at the origin, but we would like to store them compressed. To actually save space (i.e., to prevent having to store both compressed and uncompressed versions), I have created a tiny patch that adds an option called --time-only:

0.  `--time-only`

    This modifies rsync's "quick check" algorithm for finding files that need
    to be transferred, changing it from the default of transferring files with
    either a changed size or a changed last-modified time to just looking for
    files that have a changed last-modified time, ignoring size changes.  This
    is useful when remote files are uncompressed but a local copy should be
    stored compressed, togethe, with a zero-size stub to prevent re-transfers.
    This option is only useful when pulling files, as it requires post-transfer
    compression and truncation of files whilst preserving the original
    modification time of the stub. Thus the option is not transmitted to the
    remote side, also preventing the server from croaking on an unknown option.

I hope you will accept this extremely simple patch (only 3 code lines modified, two of which are declarations of the flag), which I have found very useful (saves me several terabytes of storage space!).

Sincerely,
Stein Haugan
Institute of Theoretical Astrophysics, University of Oslo


pjotrek-b:
This sounds very useful!

I'm working with similar scenarios with synchronizing large scale archival media contents across sites.

steinhh:
I may have messed up the PR:

With the release of 3.4.4, I wanted to have my changes at the tip of the new v3.4-stable b/c that's what I wanted to install on my system, but "accidentally" merged in the master first. Following advice from Gemini I replaced the previous time-only-option branch with a clean one from 3.4-stable, with force-push. I guess that's why there are conflicts.

I'm new to this game, but my guess is the cleanest way forward is to cancel this PR and create a new one? But I don't(?) want to create a pull request from a branch off v3.4-stable into master? Please advise, someone.

Btw @pjotrek-b thanks for finding & referencing this PR, do you have any suggestions for the way forward? When/if there's a new PR I guess you should update your feature request so it points the right way.

pjotrek-b:
@steinhh : Unfortunately, I'm also not part of the rsync team - and couldn't help here...
I wish you all the best! :star2:

steadytao:
> I'm new to this game, but my guess is the cleanest way forward is to cancel this PR and create a new one? But I don't(?) want to create a pull request from a branch off v3.4-stable into master? Please advise, someone.

You do not necessarily need to open another PR. The cleanest approach is to recreate the branch from current `master`, cherry-pick only the `--time-only` change and its tests then force-push that cleaned history to the existing PR branch.

Issue #499, opened 2023-07-01
feature request: support Cache Directory Tagging Specification


calestyo:
Hey.

This had already been [asked previously on the mailing list](https://lists.samba.org/archive/rsync/2013-May/028404.html) and a similar solution (`-F`) had been presented there.

However, I think it would still be nice if `rsync` could support the [Cache Directory Tagging Specification](https://bford.info/cachedir/), which basically boil down to, any directory that contains the file `CACHEDIR.TAG` with the content `Signature: 8a477f597d28d172789f06886806bc55` (and optionally some more, see the specs) is not backuped (or on the cased of `rsync` considered for copying).

Apart from avoiding possible side effects of `-F` (which AFAIU has the side effects of `--trust-sender`), the main benefit would IMO be interoperability.
Other tools (e.g. `dar` via its `--cache-directory-tagging` option) already support `CACHEDIR.TAG`, so it would be just nice to be able to use it with `rsync` to, without having to manually "re-build" it with `-F`.

Thanks for consideration,
Chris

jkt628:
it would be nice to have `--exclude-caches` and variants from GNU tar, which were added there back in 2007.

idallen:
Yes, `rsync` would benefit from obeying the `CACHEDIR.TAG` that other programs obey, such as GNU `tar`.  Saying that you can simulate the same thing by adding `rsync`-specific filter files everywhere isn't helpful since that (1) doubles the number of files one has to add everywhere, (2) opens the security risk of `--trust-sender`, and (3) only works for `rsync`.

PR #757, opened 2025-04-27
Add negation to merge files, adds `CACHEDIR.TAG` support

kaspar030:
This PR makes the previously forbidden `!` modifier available to `merge` and `dir-merge` filters.
It's effect is that it turns the default for that filter file to `- /**` (which is added *after* the merge file rules.

Additionally, unless the `e` modifier is also set, another rule for including the rules file is added.

This by itself makes it possible to support `CACHEDIR.TAG` files by specifying `--filter='dir-merge,+! CACHEDIR.TAG'`. The `CACHEDIR.TAG` is still parsed. The first line is usually `Signature: 8a477f597d28d172789f06886806bc55`, so currently a rule for that is added, but it would in practice never(tm) match.

I could imagine adding another modifier specifying a merge file is just a tag, meaning it is not parsed at all. Maybe `t`. Or, special-case the `CACHEDIR.TAG` filename. Or, ignore all rules with the exact content `Signature: 8a477f597d28d172789f06886806bc55`. But also this is already very usable as is, and IMO a pretty small change.

Additionally it would be easy to add aliases like `tar` and others have them:

```
       --exclude-caches
              Exclude contents of directories containing file CACHEDIR.TAG,
              except for the tag file itself. -> --filter='dir-merge,+! CACHEDIR.TAG'

       --exclude-caches-all
              Exclude directories containing file CACHEDIR.TAG and the file
              itself. -> --filter='dir-merge,+e! CACHEDIR.TAG'

       --exclude-caches-under
              Exclude everything under directories containing CACHEDIR.TAG
```

Let me know what you think!

Implements #499.

### DIFF
--- a/exclude.c
+++ b/exclude.c
@@ -1239,11 +1239,10 @@ static filter_rule *parse_rule_tok(const char **rulestr_ptr,
 				rule->rflags |= FILTRULE_ABS_PATH;
 				break;
 			case '!':
-				/* Negation really goes with the pattern, so it
-				 * isn't useful as a merge-file default. */
 				if (rule->rflags & FILTRULE_MERGE_FILE)
-					goto invalid;
-				rule->rflags |= FILTRULE_NEGATE;
+					rule->rflags |= FILTRULE_DEFAULT_NONE;
+				else
+					rule->rflags |= FILTRULE_NEGATE;
 				break;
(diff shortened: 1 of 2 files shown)


greyghost509:
I would also love to see this feature

PR #766, opened 2025-05-13
Fix handling of objects with many xattrs on FreeBSD

ptrrkssn:
The current code in lib/sysxattrs.c mishandles the case when there is a file/directory with many xattrs set - if the sum of the length of all the attribute names are more than 1024 it will fail. This causes problems with backups.

The fix here just makes sure we return ERANGE if we notice that the returned data is exactly the same length as the passed in buffer which will cause the code in xattrs.c to reallocate and retry with a bigger buffer.

### DIFF
--- a/lib/sysxattrs.c
+++ b/lib/sysxattrs.c
@@ -126,8 +126,17 @@ ssize_t sys_llistxattr(const char *path, char *list, size_t size)
 	unsigned char keylen;
 	ssize_t off, len = extattr_list_link(path, EXTATTR_NAMESPACE_USER, list, size);
 
-	if (len <= 0 || (size_t)len > size)
+	if (len <= 0 || size == 0)
 		return len;
+
+	if ((size_t)len >= size) {
+		/* FreeBSD extattr_list_xx() returns 'size' as 'len' in case there are                                                              
+                   more data available, truncating the output, we solve this by signalling                                                          
+                   ERANGE in case len == size so that the code in xattrs.c will retry with                                                          
+                   a bigger buffer */
+		errno = ERANGE;
+		return -1;
+	}
 
 	/* FreeBSD puts a single-byte length before each string, with no '\0'
 	 * terminator.  We need to change this into a series of null-terminted
@@ -136,7 +145,7 @@ ssize_t sys_llistxattr(const char *path, char *list, size_t size)
 	for (off = 0; off < len; off += keylen + 1) {
 		keylen = ((unsigned char*)list)[off];
 		if (off + keylen >= len) {
-			/* Should be impossible, but kernel bugs happen! */
+			/* Should be impossible, but bugs happen! */
 			errno = EINVAL;
 			return -1;
 		}


lib/sysxattrs.c
```diff
@@ -126,17 +126,26 @@ ssize_t sys_llistxattr(const char *path, char *list, size_t size)
 	unsigned char keylen;
 	ssize_t off, len = extattr_list_link(path, EXTATTR_NAMESPACE_USER, list, size);
```
rosorio, line 127: Called this way `extattr_list_link(2)` behaves like `read(2)` and return the amount of bytes there was able to write in the list buffer which is not that informative and can misleading if list has the required size.

I suggest to perform a call of `extattr_list_link(2)` with a null pointer which returns us the exact amount of bytes required in the buffer without perform any write.

```
       ssize_t off = 0, len;

       /* Check for the required size to store the extended attibutes  */
       len = extattr_list_link(path, EXTATTR_NAMESPACE_USER, NULL, size);
```

lib/sysxattrs.c
```diff
@@ -126,17 +126,26 @@ ssize_t sys_llistxattr(const char *path, char *list, size_t size)
...
 
-	if (len <= 0 || (size_t)len > size)
+	if (len <= 0 || size == 0)
 		return len;
 
+	if ((size_t)len >= size) {
```
rosorio, line 132: The equal test is not required anymore just do `if ((size_t)len > size) `

lib/sysxattrs.c
```diff
@@ -126,17 +126,26 @@ ssize_t sys_llistxattr(const char *path, char *list, size_t size)
...
+		/* FreeBSD extattr_list_xx() returns 'size' as 'len' in case there are
```
rosorio, line 133: We can make the comment simpler : "buffer is too small, behave as linux and request a bigger buffer"

lib/sysxattrs.c
```diff
@@ -126,17 +126,26 @@ ssize_t sys_llistxattr(const char *path, char *list, size_t size)
...
 
+	if ((size_t)len >= size) {
+		/* FreeBSD extattr_list_xx() returns 'size' as 'len' in case there are                                                              
+                   more data available, truncating the output, we solve this by signalling                                                          
+                   ERANGE in case len == size so that the code in xattrs.c will retry with                                                          
+                   a bigger buffer */
+		errno = ERANGE;
+		return -1;
+	}
+
```
rosorio, line 140: Now we can perform the real call to retrieve the buffer:
`
len = extattr_list_link(path, EXTATTR_NAMESPACE_USER, list, size);
`

lib/sysxattrs.c
```diff
@@ -126,17 +126,26 @@ ssize_t sys_llistxattr(const char *path, char *list, size_t size)
...
 	/* FreeBSD puts a single-byte length before each string, with no '\0'
 	 * terminator.  We need to change this into a series of null-terminted
 	 * strings.  Since the size is the same, we can simply transform the
 	 * output in place. */
 	for (off = 0; off < len; off += keylen + 1) {
```
rosorio, line 145: This for loop is quite inefficient performing multiple memmove(2) calls, what about doing it one:

```
       keylen = (unsigned char)list[0];
       memmove(list, list+1, len-1);
       list[len-1] = '\0';
       for (off = keylen; off < (len - 1); off += (keylen + 1)) {
               keylen = (unsigned char)list[off];
               list[off] = '\0';
        }
```

lib/sysxattrs.c
ptrrkssn: Yes, and no. Doing it that way has a potential to introduce a race condition if someone changes the extattr in the (very short) time between then call with NULL was done and then you actually do the real read. Doing it the way it's done now removes that race.

lib/sysxattrs.c
ptrrkssn: The problem is that if extattr_list_link(..., size) returns len==size then we don't know if we got all the data or if there is more data to read. It *might* be exactly 'size' bytes that was available there, or it might be more, but since extattr_list_link() will return just return 'size' in both cases that that will fail. Now the code after that which parses the array might return EINVAL if we just got a partial buffer - but there is a small chance that the list of extattrs is so perfectly aligned that it terminates exactly at the 'size' border - but there are still more extattrs that could have been read if the buffer was bigger...

lib/sysxattrs.c
rosorio: If you wanna cover the error you was talking in first comment (ie. changes in the xattrs key list
change between the two `extattr_list_link` calls) in a better way and avoid an error if a key is not found late, we can do the following:

```
       size_t real_xattr_len = 0;
       keylen = (unsigned char)list[0];
       memmove(list, list+1, len-1);
       list[len-1] = '\0';
       for (off = keylen; off < (len - 1); off += (keylen + 1)) {
               keylen = (unsigned char)list[off];
               list[off] = '\0';
               real_xattr_len = off + 1;
        }

        return real_xattr_len;
```

This way if the first extattr_list_link give us :
`4` with l`ist = [0x1,'A',0x1,'B']`
and the second call
`6` with `list = [0x1,'A',0x1,'B',0x1,'C']`
we will return `4` with `list = ['A',0x0,'B',0x0]`

And in a worst case if first call returns
`4` with` list = [0x1,'A',0x1,'B']`
and the second call
`4` with `list = [0x4,'K','E','Y']` which means the key was truncated
we just return `0` which means the list buffer won't be used.

rosorio:
Hi,

I think you was right, and doing a extattr_list_link() with a null buffer has no sens, mostly because is what
rsync does if first attempt fail to resize its buffer, and because your way has less edge cases to handle than mine.
So if we can agree in something like that, I'll be happy to push the patch on FreeBSD ports.


```
ssize_t sys_llistxattr(const char *path, char *list, size_t size)
{
        unsigned char keylen;
        ssize_t off, len = extattr_list_link(path, EXTATTR_NAMESPACE_USER, list, size);

        if (len <= 0 || size == 0)
                return len;

        if ((size_t)len == size) {
                /* Buffer is too small, behave like linux
                 * syscall and request a bigger buffer */
                errno = ERANGE;
                return -1;
        }

        /* FreeBSD puts a single-byte length before each string, with no '\0'
         * terminator.  We need to change this into a series of null-terminted
         * strings.  Since the size is the same, we can simply transform the
         * output in place. */
        keylen = (unsigned char)list[0];
        memmove(list, list+1, len-1);
        list[len-1] = '\0';
        for (off = keylen; off < (len - 1); off += (keylen + 1)) {
                keylen = (unsigned char)list[off];
                list[off] = '\0';
        }

        return len;
}
```

rosorio:
Patch [committed](https://cgit.freebsd.org/ports/commit/?id=558137064296eac74a8b7d9b9a370a43e97bf329) into the FreeBSD ports tree. Hope this could be merged in the rsync tree.

PR #780, opened 2025-06-18
Add detailed error information for filtered_fwrite failures

UMAHunter:
I frequently see the following error in our local 
```
2025/04/24 01:13:29 [905477] send_files(632399, /var/ms/01/002/002/tbms1s_613F1EA121BB53)
2025/04/24 01:13:29 [905477] file has vanished: "/var/ms/01/002/002/tbms1s_613F1EA121BB53"
2025/04/24 01:13:29 [905477] rsync error: errors with program diagnostics (code 13) at log.c(245) [receiver=3.1.3]
2025/04/24 01:13:29 [905477] [receiver] _exit_cleanup(code=13, file=log.c, line=245): about to call exit(13)
```
line 244 and 245 in log.c is:
```
if (ob != outbuf && fwrite(outbuf, ob - outbuf, 1, f) != 1)
	exit_cleanup(RERR_MESSAGEIO);
```
It invokes logging message, but not detailed enough to diagnose how fwrite() fails. This patch should output the errno to stderr or stdout(avoid the 'f' in the previous line) to see what fwrite() happens.


steadytao (changes requested):
Writing this diagnostic to `stdout` when `stderr` fails looks risky; stdout may be normal output or protocol data. I think this should avoid the alternate stream fallback.

PR #781, opened 2025-06-22
Expand xattr tests for RsyncProject/rsync#766

rosorio:
When a file have many xattrs that the sum of all
attribute names is larger than 1024, rsync transfer will fail.

This issue was fixed by #766


rosorio:
Apparently cygwin doesn't support long xattrs so for now skip the test for this platform

PR #796, opened 2025-08-23
improve the attribute parsing by reducing to one the memmove for FreeBSD

rosorio:
calls when converting FreeBSD attribute list to the format used by Linux and extend the rsync unit tests for long xattrs lists.

This is part of the patches currently applied to rsync on FreeBSD


sergejsuhoveckij56-dot (changes requested):
Дубликат #

PR #800, opened 2025-08-28
Harden systemd service with safe defaults

LaserEyess:
See commit message for an overview, but this is an RFC, I can understand if the project does not want this. The goal here is defense in depth. If there is a bug in any access control or authentication mechanism in rsync, there should be additional protections to not compromise a system running `rsync --daemon` as root. This can be accomplished through systemd, so that the damage done by a hypothetical RCE is minimized as much as possible. 

**I have personally tested this and it does not affect file transfers for either unit**. That being said, I understand a lot of the options may be confusing. I can explain why each one of these is a safe default that *does not affect file transfers*.

### DIFF
--- a/packaging/systemd/rsync.service
+++ b/packaging/systemd/rsync.service
@@ -25,8 +25,30 @@ Restart=on-failure
 
 ProtectSystem=full
 #ProtectHome=on|off|read-only
+
+# These are general hardening parameters that should not affect file access
 PrivateDevices=on
 NoNewPrivileges=on
+MemoryDenyWriteExecute=on
+LockPersonality=on
+ProtectClock=on
+ProtectControlGroups=on
+ProtectHostname=on
+ProtectKernelLogs=on
+ProtectKernelModules=on
+ProtectKernelTunables=on
+ProtectProc=invisible
+ProcSubset=pid
+RestrictNamespaces=on
+RestrictRealtime=on
+SystemCallArchitectures=native
+SystemCallErrorNumber=EPERM
+SystemCallFilter=@system-service
+
+# We only listen on TCP sockets
+SocketBindAllow=ipv4:tcp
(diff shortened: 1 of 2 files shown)


LaserEyess:
@tridge I read your blog about wanting to increase defense-in-depth on rsync. I believe this PR is complementary to that goal. Further improvements could be made, such as not running rsync daemon as root. But that would be a breaking change so I left that out for this.

LaserEyess:
Marking this as ready for review since there has been no negative feedback.

steadytao:
Just a note that drafts are ordinarily not read, they are explicitly that, drafts.

However; these cannot be described as universally safe defaults without testing representative writable modules, /tmp-backed modules/temp dirs, chroot, xattrs/ACLs, socket activation and the supported systemd versions. PrivateTmp alone can alter configured file visibility. Perhaps best as an optional hardening drop-in first.

LaserEyess:
Fair enough about `PrivateTmp`. I wouldn't normally consider copying *to* a remote system's `/tmp` but I suppose that's not an impossible use case. 

As for the other point, the tests, it doesn't look like a systemd service test exists. I only briefly glanced at `testsuite/`, so maybe I'm missing something. In general testing systemd units specifically is not so easy, though with containers/VMs it's not impossible. Is there any way you would like me to approach that?

steadytao:
For testing, a small systemd integration script or documented VM/container matrix would be enough initially:
- systemd-analyze verify both units;
- start the standalone daemon and perform an upload/download against a writable module;
- repeat through socket activation;
- exercise a chrooted module and xattrs/ACLs where supported;
- confirm configured module paths and temporary directories remain accessible.

Debian or Ubuntu plus Fedora or Alma would cover the main systemd variation. If that passes, I would be more comfortable with the remaining directives left as defaults rather than an optional drop-in.

PR #816, opened 2025-10-02
fix: make `--info=progress2` consistently display time-remaining

### LINKED ISSUE #392: Confusing output with --info=progress2
# The problem

Unlike `--info=progress`, `--info=progress2` does not insert newlines after each transferred file.

This is a problem because `progress2` displays 2 different kinds of information:
- During file transfer it seems to display instantaneous transfer rate and ETA (+ other things)
- Between 2 file transfers it seems to display average transfer rate since the beginnning and time elapsed (+ other things).

Without newlines, these 2 overwrite each other, making it hard to read, and confusing people (see comments on https://unix.stackexchange.com/questions/215271/understanding-the-output-of-info-progress2-from-rsync).

A quick fix is to use something like `--info=name,progress2`, which causes newlines to be inserted, thus separating the 2 kinds of output.

# Possible solutions
## Newlines
Adding newlines to `--info=progress2`.
Is it really meaningful to have one line per file when asking for total progress though ?

## Simpler output
Only display instantaneous transfer rate and ETA (+ other things), as it seems to be the most relevant information.
This would have the advantage of keeping the one line output, while also making it readable.

## Other solutions
- Adding more output control by extending `--out-format` to such information.
- Always displaying these 4 metrics instead of alternating.
- ...

tinnamchoi:
Implements https://github.com/RsyncProject/rsync/issues/392#issuecomment-1537286136:
> I plan to make `--info=progress2` switch to a consistent time-remaining format for all the output lines since it is overwriting the line.

#### Previous Behaviour

During file each transfer: Total time-remaining (i.e. ETA) is displayed.
After file each transfer: Total time taken is displayed, which gets overwritten unless using something like `--info=name` to insert a new line.

#### New Behaviour

Total time-remaining is always displayed.
`--info=progress` unchanged.

This is particularly significant when transfering a large number of small files, where it becomes difficult to see the overall time-remaining. 

Potentially resolves #392


steadytao (changes requested):
This should probably have a regression test. In `progress2`, `ofs` can be less than `size` at the final update so this may show remaining time at the end of the run instead of elapsed time.

PR #835, opened 2025-12-30
reject negative token values in compressed stream receivers

tridge:
Validate that token numbers read from compressed streams are non-negative. A negative token value would cause the return value of recv_*_token() to become positive, which callers interpret as literal data length, but no data pointer is set on this code path.

While this only causes the receiver to crash (which is process-isolated and only affects the attacker's own connection), it's still undefined behavior.

Reported-by: Will Sergeant <wlsergeant@gmail.com>


willsarg:
Thank you for looking into this!
-Will Sergeant

Issue #839, opened 2026-01-07
--checksum time out waiting for file list


tenox7:
When using `--checksum` option I get timeouts while waiting for server to generate file list. Both recursive and `--no-i-r`.

Even with `--timeout=xxx` far exceeding file list generation time, both client and server end up stuck at `pselect6()` forever.

I have tried `SO_KEEPALIVE` and all combinations of partial/whole recursive, non-recursive, etc etc. Nothing really helps.

This seems quite widely reported issue "on the internet" with most popular workarounds ranging from never using `--checksum` to splitting it to smaller subdirectories.

tenox7:
Submitting an experimental fix #840

PR #840, opened 2026-01-07
send keepalives during checksum operations in daemon mode

tenox7:
This is an experimental PR that helps to prevent timeouts during file list generation with `--checksum`. 

Even with `--timeout=xxx` far exceeding file list generation time, both client and server end up stuck at `pselect6()` forever. I have tried `SO_KEEPALIVE` and all combinations of partial/whole recursive, non-recursive, etc etc. Nothing really helps.

I have NOT extensively tested it, but it resolves my issue #839.

### DIFF
--- a/checksum.c
+++ b/checksum.c
@@ -36,11 +36,30 @@
 #endif
 
 extern int am_server;
+extern int am_daemon;
+extern int always_checksum;
 extern int whole_file;
 extern int checksum_seed;
 extern int protocol_version;
 extern int proper_seed_order;
 extern const char *checksum_choice;
+extern void maybe_send_keepalive(time_t now, int flags);
+
+static time_t last_checksum_keepalive = 0;
+
+static inline void send_checksum_keepalive(void)
+{
+	time_t now;
+
+	if (!am_daemon || !always_checksum)
+		return;
+
+	now = time(NULL);
+	if (now - last_checksum_keepalive >= 5) {
+		last_checksum_keepalive = now;
+		maybe_send_keepalive(now, MSK_ALLOW_FLUSH);
+	}
+}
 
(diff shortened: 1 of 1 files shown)


dr-who:
@tenox7 I reproduced #839 from your description and dug into the mechanism — your diagnosis (keepalives during the checksum pass) is right. Two things I found while testing this patch, which I thought were worth passing on:

**1. The 5-second interval means it doesn't fix the case in the report.** I applied this PR and re-ran a reproducer (single 2 GiB file in a daemon module, ~4.3s to hash):

| build | `--timeout=3` | `--timeout=6` | `--timeout=12` |
|---|---|---|---|
| master | times out | times out | times out |
| this PR | **still times out** | fixed | fixed |

With `--timeout=3` the receiver gives up at 3s, before `now - last_checksum_keepalive >= 5` ever fires. Any `--timeout` below ~6 stays broken.

**2. The `am_daemon` gate is narrower than the bug.** It also happens over a remote shell — master fails with `--checksum --timeout=2` and `--timeout=1` there.

The good news is rsync already has exactly the right primitive: `maybe_send_keepalive()` self-throttles on `allowed_lull`, which `set_io_timeout()` computes as **half of `--timeout`** — so it adapts to whatever timeout the user set instead of a fixed 5s, and the generator/sender/receiver already use it for their long stretches.

I've put that up as #1031, with the tick folded into a single loop shared by all six digests (this file has six copies of the same chunk loop, which is why this PR had to patch each one). It's verified digest-identical to master in both directions across the chunk-boundary sizes. Full credit to you for the report and for spotting that keepalives were the answer — please take anything useful from it, or say the word if you'd rather fold the interval change into this PR instead and I'll close mine.

PR #861, opened 2026-03-06
rsyncd.conf: document the purpose of the temp_dir parameter

### LINKED ISSUE #820: Rsyncd has a temp dir feature that is undocumented
Hi,

Using rsync version 3.2.7, I stumbled upon a `temp dir` option in the **rsyncd** configuration. 
If I set under a module the temp dir like so:
```
[data]
  path = /data
  temp dir = /.rsyncd_tmp
```
Provided that the /data/.rsyncd_tmp directory is created, transfers to the rsync module `data` will use /data/.rsyncd_tmp as a temporary directory during the transfer.

I see references to that option in the [daemon-parm.txt](https://github.com/RsyncProject/rsync/blob/797e17fc4a6f15e3b1756538a9f812b63942686f/daemon-parm.txt#L46) file as well.

However, there is no reference to that feature in the man page rsyncd.conf.5

It seems to work fine so is that just missing documentation or should that feature not be used?

devZer0:
Add a description explaining that temp_dir can also configured in rsyncd.conf

fixes https://github.com/RsyncProject/rsync/issues/820

### DIFF
--- a/rsyncd.conf.5.md
+++ b/rsyncd.conf.5.md
@@ -1073,6 +1073,16 @@ in the values of parameters.  See that section for details.
     **system()** call's default shell), and use RSYNC_NO_XFER_EXEC to disable
     both options completely.
 
+0.  `temp dir`
+
+    Specifies a directory that rsync should use for temporary files created
+    during the transfer of updated files. If that directory is on a different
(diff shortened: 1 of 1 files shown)


tridge:
fixed some minor errors, thanks!

Issue #715, opened 2025-01-16
rsync: failed verification -- update discarded - regression from CVE fixes


samueloph:
Hello,

We've got the following bug report on Debian:
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1093160

Then I was able to repro this regression on both Debian and Amazon Linux.

To repro, you need a host you can ssh to, in this case I'm using localhost:
```
mkdir "$HOME"/real-dir
ln -s "$HOME"/real-dir "$HOME"/dir
mkdir /tmp/dir
echo "foo" > /tmp/dir/file
rsync -KRlptzv /tmp/dir/file localhost:
touch /tmp/dir/file
touch dir/file
rsync -KRlptzv /tmp/dir/file localhost:
```

This results in:
```
rsync -KRlptzv dir/file localhost:
dir/
dir/file
WARNING: dir/file failed verification -- update discarded (will try again).
dir/file
ERROR: dir/file failed verification -- update discarded.

sent 146 bytes  received 165 bytes  207.33 bytes/sec
total size is 4  speedup is 0.01
rsync error: some files/attrs were not transferred (see previous errors) (code 23) at main.c(1179) [sender=3.1.2]
```

samueloph:
cc @ncopa so they can check if Alpine is also affected.

carnil:
If I have not done an error on bisecting the issue it is introduced with c35e28331f10ba6eba370611abd78bde32d54da7

yankov-pt:
@carnil Did you use the above snippet to reproduce? 
I've tried reproducing with bisecting and I keep triggering this behavior going back to at least several versions. 
With the oldest binary on a system I have - `3.1.3` still happens.

vinc17fr:
BTW, when testing, do not forget `--rsync-path=/path/to/tested/rsync` if the version to be tested is not first in `$PATH`.

carnil:
@yankov

> [@carnil](https://github.com/carnil) Did you use the above snippet to reproduce? I've tried reproducing with bisecting and I keep triggering this behavior going back to at least several versions. With the oldest binary on a system I have - `3.1.3` still happens.

I'm using the following (but it does not yet work for automatic bisection, so did manual test)
```bash
#!/bin/sh

if
    rm -rf -- "$HOME/real-dir" "$HOME/dir" "/tmp/dir"
    git clean -d -x -f && ./configure --with-included-zlib=no --disable-md5-asm --with-rrsync && make
then
[...]

PR #864, opened 2026-03-15
receiver: fix basis file open for directory symlinks (issue #715)

samueloph:
(The patch on this PR was generated with Claude)

I've noticed that the latest two commits were done with the use of an AI agent, I took that as a signal that it would be ok to submit this PR.

The regression from #715 has been bothering users for a roughly one year now, and since I don't have enough time to investigate the source code and handwrite a patch myself, I though I would give Claude a chance.

I've reviewed everything in this PR myself, but I didn't go too deep in understanding the internals of rsync and that's where there's room for issues, I've focused heavily on unit tests to attest behaviors.

For something sensitive like this, I guess the only way to be sure is to have @tridge double check it, but I will understand if you don't want to review it. In the worst case, I suppose this PR is better than nothing.

These is a summary of the things I did to lower the chances of hallucinations and mistakes:

(1) Ground the agent in primary sources - fetch the issue, read the CVE fix commit and full function before proposing changes.
(2) Require build + test at every step - the agent compiled, wrote a regression test, verified it fails without the fix, and ran the full suite after each change. 
(3) Explicitly ask the agent to verify security properties - without prompting, the agent hadn't analyzed whether the fix reintroduces the CVE. 
(4) Explicitly ask about flag interactions - this uncovered two bugs (--partial-dir with protocol < 29, stale basedir in fallback path) the agent missed on first pass.
(5) Demand thoroughness after mistakes - after the agent missed those two, pushing harder uncovered a third bug (--backup xattr handling via pointer identity).
(6) Add test coverage for each discovered bug - the final test has 8 scenarios covering every edge case found.
(7) Test the absence of a test - the original CVE fix had no regression test, so one was added and validated by temporarily reverting the fix. The core lesson: the agent's first-pass fix was correct for the reported bug but had three latent issues causing silent data loss in edge cases; each was found only after explicit human direction to look harder.
(8) Explicitly ask it to verify how this patch interacts with the fixes for CVE-2024-1208{4..8}, which is where the regression came from.

### DIFF
--- a/receiver.c
+++ b/receiver.c
@@ -753,6 +753,7 @@ int recv_files(int f_in, int f_out, char *local_name)
 			  && check_filter(&daemon_filter_list, FLOG, fnamecmp, 0) < 0)) {
 				fnamecmp = fname;
 				fnamecmp_type = FNAMECMP_FNAME;
+				basedir = NULL;
 			}
 		} else {
 			/* Reminder: --inplace && --partial-dir are never
@@ -768,6 +769,19 @@ int recv_files(int f_in, int f_out, char *local_name)
 				fnamecmp = fname;
 		}
 
+		/* For FNAMECMP_FNAME, split into basedir (dirname) + fnamecmp
+		 * (basename) so that secure_relative_open() follows symlinks
+		 * in the directory part. This is needed because the receiver's
+		 * destination may legitimately contain directory symlinks
+		 * (e.g. when using -K/--copy-dirlinks). The basedir is opened
+		 * following symlinks, while the file component uses O_NOFOLLOW.
+		 * See: https://github.com/RsyncProject/rsync/issues/715 */
+		if (fnamecmp_type == FNAMECMP_FNAME && basedir == NULL
+		 && fnamecmp == fname && file->dirname && !local_name) {
+			basedir = file->dirname;
+			fnamecmp = (char*)file->basename;
+		}
+
 		/* open the file */
 		fd1 = secure_relative_open(basedir, fnamecmp, O_RDONLY, 0);
 
@@ -775,6 +789,7 @@ int recv_files(int f_in, int f_out, char *local_name)
 			if (fnamecmp != fname) {
 				fnamecmp = fname;
 				fnamecmp_type = FNAMECMP_FNAME;
+				basedir = NULL;
 				fd1 = do_open_nofollow(fnamecmp, O_RDONLY);
 			}
 
@@ -792,6 +807,12 @@ int recv_files(int f_in, int f_out, char *local_name)
 			// path name as a single string
 			pathjoin(fnamecmpbuf, sizeof fnamecmpbuf, basedir, fnamecmp);
 			fnamecmp = fnamecmpbuf;
+			/* For FNAMECMP_FNAME, the reconstructed path is
+			 * identical to fname. Restore the pointer so that
+			 * downstream pointer comparisons (e.g. in
+			 * finish_transfer) continue to work. */
+			if (fnamecmp_type == FNAMECMP_FNAME)
+				fnamecmp = fname;
 		}
 
 		one_inplace = inplace_partial && fnamecmp_type == FNAMECMP_PARTIAL_DIR;
--- a/testsuite/secure-relative-open.test
+++ b/testsuite/secure-relative-open.test
@@ -1,0 +1,111 @@
+#!/bin/sh
+
+# Regression test for the CVE fix in commit c35e283 ("receiver: use
+# secure_relative_open() for basis file").
+#
+# secure_relative_open() enforces O_NOFOLLOW on every path component of
+# the relpath, preventing the receiver from following symlinks when
+# opening basis files for delta comparison. This stops a malicious
+# sender from causing the receiver to read files outside the
+# destination tree via crafted symlinks.
+#
+# This test verifies that when a basis directory (--copy-dest) contains
+# a symlink as an intermediate path component, the receiver does NOT
+# follow it to open the basis file. The generator may find the file
+# (it uses link_stat/do_open_checklinks which follow intermediate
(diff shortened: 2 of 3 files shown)


mcepl:
Was this supposed to fix `-K` option? [Because it didn’t](https://build.opensuse.org/package/show/home:mcepl/rsync). This [backup script](https://github.com/user-attachments/files/26291258/backup.sh) leads to 
this [backup-log.txt](https://github.com/user-attachments/files/26291263/backup-log.txt), which contains precious pearls like:

```
WARNING: .config/dotfiles/weechat/python/matrix/_weechat.py failed verification -- update discarded (will try again).
WARNING: .config/dotfiles/weechat/python/matrix/bar_items.py failed verification -- update discarded (will try again).
WARNING: .config/dotfiles/weechat/python/matrix/buffer.py failed verification -- update discarded (will try again).
WARNING: .config/dotfiles/weechat/python/matrix/colors.py failed verification -- update discarded (will try again).
WARNING: .config/dotfiles/weechat/python/matrix/commands.py failed verification -- update discarded (will try again).
WARNING: .config/dotfiles/weechat/python/matrix/completion.py failed verification -- update discarded (will try again).
WARNING: .config/dotfiles/weechat/python/matrix/config.py failed verification -- update discarded (will try again).
WARNING: .config/dotfiles/weechat/python/matrix/globals.py failed verification -- update discarded (will try again).
WARNING: .config/dotfiles/weechat/python/matrix/message_renderer.py failed verification -- update discarded (will try again).
WARNING: .config/dotfiles/weechat/python/matrix/server.py failed verification -- update discarded (will try again).
WARNING: .config/dotfiles/weechat/python/matrix/uploads.py failed verification -- update discarded (will try again).
WARNING: .config/dotfiles/weechat/python/matrix/utf.py failed verification -- update discarded (will try again).
WARNING: .config/dotfiles/weechat/python/matrix/utils.py failed verification -- update discarded (will try again).
.config/dotfiles/weechat/python/matrix/_weechat.py
.config/dotfiles/weechat/python/matrix/bar_items.py
.config/dotfiles/weechat/python/matrix/buffer.py
.config/dotfiles/weechat/python/matrix/colors.py
.config/dotfiles/weechat/python/matrix/commands.py
.config/dotfiles/weechat/python/matrix/completion.py
.config/dotfiles/weechat/python/matrix/config.py
.config/dotfiles/weechat/python/matrix/globals.py
.config/dotfiles/weechat/python/matrix/message_renderer.py
.config/dotfiles/weechat/python/matrix/server.py
.config/dotfiles/weechat/python/matrix/uploads.py
.config/dotfiles/weechat/python/matrix/utf.py
.config/dotfiles/weechat/python/matrix/utils.py
ERROR: .config/dotfiles/weechat/python/matrix/_weechat.py failed verification -- update discarded.
ERROR: .config/dotfiles/weechat/python/matrix/bar_items.py failed verification -- update discarded.
ERROR: .config/dotfiles/weechat/python/matrix/buffer.py failed verification -- update discarded.
ERROR: .config/dotfiles/weechat/python/matrix/colors.py failed verification -- update discarded.
ERROR: .config/dotfiles/weechat/python/matrix/commands.py failed verification -- update discarded.
ERROR: .config/dotfiles/weechat/python/matrix/completion.py failed verification -- update discarded.
ERROR: .config/dotfiles/weechat/python/matrix/config.py failed verification -- update discarded.
ERROR: .config/dotfiles/weechat/python/matrix/globals.py failed verification -- update discarded.
ERROR: .config/dotfiles/weechat/python/matrix/message_renderer.py failed verification -- update discarded.
ERROR: .config/dotfiles/weechat/python/matrix/server.py failed verification -- update discarded.
```
and ends with
```
rsync error: some files/attrs were not transferred (see previous errors) (code 23) at main.c(1338) [sender=3.4.1]
pá 27. března 2026, 01:20:01 CET: Rsync failed for -K with status 23
```

samueloph:
@mcepl thanks for testing, can you confirm that you do have a patched rsync both on the sender and receiver side (although I believe you just need the receiver to be patched)?

The patch is not there in https://build.opensuse.org/package/show/home:mcepl/rsync

mcepl:
> @mcepl thanks for testing, can you confirm that you do have a patched rsync both on the sender and receiver side (although I believe you just need the receiver to be patched)?
> 
> The patch is not there in https://build.opensuse.org/package/show/home:mcepl/rsync

Correct, the fixed `rsync` must be on both sides. With rsync from these packages https://build.opensuse.org/package/show/home:mcepl/rsync on client (Tumbleweed) and server (15.6), there are no warnings and errors for me!

samueloph:
Awesome, thanks for checking!

kraptor:
We are tracking this patch and issue in openSUSE here:
* https://bugzilla.opensuse.org/show_bug.cgi?id=1236359

I'll make sure we patch (or update the package to a new release including the patch) in the official package as soon as the fix is merged. Thanks for testing  @mcepl!!

samueloph:
I've uploaded this patch to rsync in Debian Experimental, it's in version `3.4.1+ds1-8~exp1`, it will be available in a few hours from now.

Once our CI kicks-in, we should get some results in https://qa.debian.org/excuses.php?package=rsync&experimental=1. Debian's CI runs against all reverse-dependencies of rsync.

tridge:
@samueloph it is extremely hard to prove that this doesn't re-introduce the security issue that the original change fixed.
I think what we're going to need to do is use openat2() with RESOLVE_BENEATH for more recent linux kernels (5.6 and later?) and use O_RESOLVE_BENEATH on recent freebsd/openbsd. Are you interested in developing a PR that does that? That would directly address the issue as it would restrict symlink resolution to be within a directory tree, which is a very good match for the semantics we need and presumably is race free (at least the kernel has a chance of making this race free)

tridge:
fixed by #887

PR #867, opened 2026-04-05
Fix glibc-2.43 constness warnings

hhoffstaette:
Glibc 2.43 added C23 const-preserving overloads to various string functions, which change the return type depending on the constness of the argument(s). Currently this leads to warnings from calls to strtok() or strchr() which can break the build e.g. when -Werror is used. Fix this by properly declaring the respective variable types.

### DIFF
--- a/access.c
+++ b/access.c
@@ -99,7 +99,7 @@ static void make_mask(char *mask, int plen, int addrlen)
 	return;
 }
 
-static int match_address(const char *addr, const char *tok)
+static int match_address(const char *addr, char *tok)
 {
 	char *p;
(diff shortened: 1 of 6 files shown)


hhoffstaette:
zlib on MacOS won't build with -std=gnu23 because of old-style protoypes, but that's a zlib problem. It only now surfaces because (presumably) the Homebew action already has autoconf-2.73, which tries to force C23 and therefore makes clang reject old prototypes.
See e.g. https://bugs.gentoo.org/972320

tridge:
@hhoffstaette thanks!

PR #894, opened 2026-05-19
add fileflags option for BSD and others

tridge:
This is based on the fileflags.patch in the rsync-patches archive
needs a bit more work before merging

todo:
 - [ ] --force option is removed, breaking existing usage
 - [ ] need to carefully look at security implications
 - [x] possibly support on linux with chattr (ie. EXT2_IOC_SETFLAGS or similar)
 - [x] need to change path based calls to fd based calls, with RESOLVE_BENEATH


daeho-ro:
Is this PR active? Just in case.

PR #901, opened 2026-05-20
Set defines by OS name if cross-compiling, especially for Android

ribbons:
If `MKNOD_CREATES_FIFOS` is not defined by configure (such as when cross-compiling), `mkfifo()` is available but `mkfifoat()` is not then compilation fails since 30656c5e35.

(I discovered this when trying to cross-compile v3.4.3 for Android, targeting API level 21 which provides `mkfifo()` but not `mkfifoat()`).

This PR adds a configure function check for `mkfifoat` and updates the preprocessor conditional that wraps the call to it in syscall.c to reference `HAVE_MKFIFOAT`.

### DIFF
--- a/.github/workflows/android-static-build.yml
+++ b/.github/workflows/android-static-build.yml
@@ -63,13 +63,8 @@ jobs:
 
           # Bionic doesn't declare lchmod()/lutimes() until API 36, but the
           # symbols link, so configure mis-detects them -- force them off so
-          # rsync uses its fallbacks. The other cache vars restore values
-          # that configure can't probe when cross-compiling (Android runs a
-          # normal Linux kernel, so these match the native Linux result).
-          export ac_cv_func_lchmod=no ac_cv_func_lutimes=no \
-                 rsync_cv_HAVE_SOCKETPAIR=yes \
-                 rsync_cv_MKNOD_CREATES_FIFOS=yes \
-                 rsync_cv_MKNOD_CREATES_SOCKETS=yes
+          # rsync uses its fallbacks.
+          export ac_cv_func_lchmod=no ac_cv_func_lutimes=no
 
           # Self-contained build: drop optional external libraries so the
           # static binary needs nothing at runtime. rsync keeps md5/md4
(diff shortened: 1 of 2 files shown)


tridge:
@ribbons I'm planning to have android builds as part of the main rsync repo, both normal builds and specific termux deb pkgs. Are you interested in helping with that? I'd like android to be tested in CI to ensure it doesn't bitrot

ribbons:
@tridge That sounds great!  Sure thing - happy to help (though with the caveat that I've not used Termux in anger before)!

If you're happy to add another third-party action, I've had pretty good results using [
ReactiveCircus/android-emulator-runner](https://github.com/ReactiveCircus/android-emulator-runner) within Actions and I'm guessing with some fiddling it _should_ then be possible to run the test suite within the emulator as part of the CI.

On the topic of this PR, do you think this is the right approach or is it getting the correct result the wrong way?  I was a bit on the fence before but seeing your workaround in 88393140250b491463bf52900abb6c7bac5ee154 made me think it might be better to adjust configure.ac to automatically set HAVE_SOCKETPAIR, MKNOD_CREATES_FIFOS and MKNOD_CREATES_SOCKETS when cross-compiling and the target is android ?  That would then both fix this specific issue and give the correct feature detection for anyone downstream performing builds.

tridge:
@ribbons I'm leaning towards the auto-set of options, but I'd be happy to declare you as being the rsync on android maintainer and let you decide on the approach. The aim would be that rsync.samba.org downloads has usable android builds for the various architectures and environments that android uses, maybe in an android subdirectory with a readme explaining when to use which one and how to install.
Note that I've found claude quite useful in setting up these CI jobs.

ribbons:
@tridge great, thanks - I've switched this PR to that approach instead and so `HAVE_C99_VSNPRINTF`, `HAVE_SECURE_MKSTEMP`, `HAVE_SOCKETPAIR`, `MKNOD_CREATES_FIFOS` and `MKNOD_CREATES_SOCKETS` should all now get defined automatically when cross-compiling for Android.  I took the liberty of making `HAVE_SOCKETPAIR`, `MKNOD_CREATES_FIFOS` and `MKNOD_CREATES_SOCKETS` set themselves when cross-compiling for _any_ Linux target as unless I'm mistaken it looks like they should apply to anything Linux.

PR #902, opened 2026-05-20
Fix chmod test suite issue on older systems

tridge:
On systems without O_RESOLVE_BENEATH or openat2/RESOLVE_BENEATH the test suite would fail unless the system was specifically skipped in the chmod-symlink-race test. This fixes the issue and broadens the test coverage to all systems.
This fixes a testsuite issue reported on HPE NonStop

### DIFF
--- a/t_chmod_secure.c
+++ b/t_chmod_secure.c
@@ -17,6 +17,11 @@
 
 #include <sys/stat.h>
 
+#ifdef __linux__
+#include <sys/syscall.h>
+#include <linux/openat2.h>
+#endif
+
 int dry_run = 0;
 int am_root = 0;
 int am_sender = 0;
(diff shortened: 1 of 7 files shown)


t_chmod_secure.c
```diff
@@ -30,6 +35,42 @@ short info_levels[COUNT_INFO], debug_levels[COUNT_DEBUG];
...
+		how.flags = O_RDONLY | O_DIRECTORY;
+		how.resolve = RESOLVE_BENEATH | RESOLVE_NO_MAGICLINKS;
+		fd = syscall(SYS_openat2, AT_FDCWD, ".", &how, sizeof how);
+		if (fd >= 0) {
+			close(fd);
+			return 1;
+		}
+		/* ENOSYS = kernel < 5.6.  Fall through to the O_RESOLVE_BENEATH
+		 * probe in case we're a Linux build running on a kernel that
+		 * gained O_RESOLVE_BENEATH via some out-of-tree backport. */
```
xokdvium, line 61: A bit of a drive-by comment, but O_RESOLVE_BENEATH isn't a Linux thing and would never be defined. Nothing like this has been mainlined, so there's nothing to "backport".

Did you have a concrete example of people actually doing this with some hackjobs?

From the manpage:

> FreeBSD's O_RESOLVE_BENEATH semantics are based on
       Linux's RESOLVE_BENEATH and the two are now functionally
       equivalent.

BSD's O_RESOLVE_BENEATH was modelled after Linux's RESOLVE_BENEATH, so it's doubly confusing https://reviews.freebsd.org/D25886

PR #918, opened 2026-05-25
Windows native support (with Microsoft Visual Studio)

SamuelMarks:
```cmd
> cmake -G "Visual Studio 18 2026" -DCMAKE_TOOLCHAIN_FILE="..\vcpkg\scripts\buildsystems\vcpkg.cmake" -A x64 -S . -B build
-- Selecting Windows SDK version 10.0.26100.0 to target Windows 10.0.26200.
-- The C compiler identification is MSVC 19.50.35730.0
[… omitted for brevity]
> cmake --build build --config Debug --parallel
[… omitted for brevity]
```

Which will produce a neat 2MB rsync.exe. Release mode gives a 440KB rsync.exe.


tridge (changes requested):
thanks!
Please add a CI job as well to test this, preferably with it running the test suite

devZer0:
@SamuelMarks wo, we really get a native windows rsync port, i.e. we can really run this without cygwin or WSL? 

but wouldn't this need some little description/announcement or further information , what's being added here and in which way it probably differs from unix rsync? 

what about path names ("\" vs "/") , drive letters and such ?

tridge:
@SamuelMarks where is the auto-win-msvc shim layer referenced by config.h.in?
also, please avoid adding cmake

SamuelMarks:
@tridge https://github.com/SamuelMarks/auto-win-msvc

Bringing rsync (and other ported projects) into the CI; so when that passes we should have some degree of confidence.

@devZer0 still plenty of testing to be done; particularly the introduction of new tests. Lots of Windows stuff is annoying (UTF 16 with BOM on paths that need allocation; UNC paths ; etc.).

tridge:
@SamuelMarks please avoid cmake

PR #909, opened 2026-05-22
fixes for android/termux setup

tridge:

PR #920, opened 2026-05-26
Run test suite in emulator as part of Android build workflow

ribbons:
@tridge - inspired by our initial exchange in #901 here is a proof of concept of running the test suite within the Android emulator under Termux as part of the Android build workflow you set up previously.  The script can also be run locally if a build targeting Android has been done and an Android emulator is running (I've made the script skip installing Termux or Python if they are already installed to make it fairly quick to keep running the tests).

~~I need to have a few more passes through the script to make things cleaner and tidier, add some more checks, write a decent commit message etc but I'm happy to chip away at all that stuff over the next few days if this looks like the kind of approach you'd be in favour of?~~

~~As naming things is always difficult, do you think `android-run-tests` is a logical / consistent name for the script and is under `support` a sensible place for it (very happy to change if you have strong feelings otherwise)?~~

### DIFF
--- a/.github/workflows/android-static-build.yml
+++ b/.github/workflows/android-static-build.yml
@@ -1,13 +1,11 @@
-name: Build static rsync for Android
+name: Test rsync on Android
 
 # Cross-compiles statically-linked rsync binaries with the Android NDK,
 # suitable for dropping onto a phone (adb push / Termux) with no shared
 # libraries. arm64-v8a covers all modern phones; armeabi-v7a covers older
 # 32-bit devices. The binaries are uploaded as workflow artifacts.
 #
-# These are cross-compiled, so the test suite can't run here; we sanity
-# check that each binary is the right architecture, is static, and that
-# it executes (`--version`) under qemu-user.
+# Run the test suite within Termux under the Android Emulator.
 
 on:
   push:
@@ -43,6 +41,8 @@ jobs:
           - abi: armeabi-v7a     # older 32-bit phones
             triple: armv7a-linux-androideabi
             qemu: qemu-arm-static
+          - abi: x86_64
+            triple: x86_64-linux-android
     steps:
       - uses: actions/checkout@v4
         with:
@@ -89,6 +89,7 @@ jobs:
           "$STRIP" rsync
 
       - name: Verify binary
+        if: matrix.abi != 'x86_64'
         shell: bash
         run: |
           set -euo pipefail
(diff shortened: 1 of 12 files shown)


ribbons:
Okay, have mulled this over more and done a bunch of tidying up.  Have moved the script from support to `test-android` at the top level to match other Makefile helper scripts.  Instead of the script calling make and then running the tests, have adjusted the check target so that it automatically runs the helper script if the build has been configured for cross-compilation and `host_os` matches `*android*`.

Also made a bunch of smaller things tidier:
* ~~`CHECK_PROGS` are passed to the helper script for the tar command so that it doesn't
contain a hard-coded list which needs to be kept in sync.~~
* `CHECK_J` is passed through from the Makefile to `runtests.py` to match running the tests natively.
* The files are piped into the emulator directly instead of using a temporary tar file.
* Removed the need for `adb root` as part of setting up the tests.

A few points I probably should have explained when I first opened this PR (but ran out of time):
* I added x86_64 to the build matrix and the tests are currently only run on that architecture as the x86_64 emulator on an x86_64 runner gives (fairly) decent performance compared to emulating ARM on x86_64.
* The x86_64 run is failing because there are tests that currently fail under Android.  I'll work my way through and see which ones are due to bad environment setup on my part vs actual bugs that the test suite is picking up.
* I've pinned the android-emulator-runner version by commit hash [as recommended by GitHub for security reasons](https://docs.github.com/en/actions/reference/security/secure-use#using-third-party-actions) but the downside of this is that it will need to be bumped manually from time to time (or semi-automatically by something like Dependabot).  If you'd prefer a different approach, let me know.
* The helper script automatically downloads the latest version of Termux from their GitHub releases if needed (as well as the python package), as this is only run within the emulator this seems like it should be a manageable risk but again something to be aware of.

Would appreciate some feedback @tridge if you are able - brickbats or bouquets equally welcomed :smile:

ribbons:
All tests should now be passing (or automatically skipping) - a bunch of the failures were due to my only pulling across a subset of the files to the emulator so that was easily remedied.  The others were due to the hard links issue you already discovered so I've cherry picked a couple of your commits from #909 and added a commit of my own to skip the hardlinks section of `alt-dest-deep` and to automatically skip `hardlinks-deep`.

It would probably also be worth adding some different emulator versions to extend the coverage at some point but I'd suggest that's probably best left to a follow-up PR to avoid making this one too much of a monster?

Issue #928, opened 2026-05-29
rsync over ssh with relative basis --link-dest=../snap.1 can fail with "failed verification -- update discarded" in 3.4.3


JetAppsClark:
## Title

`--link-dest=../snap.1` over SSH can fail with `failed verification -- update discarded` in rsync 3.4.3

## Summary

I am seeing reproducible `failed verification -- update discarded` errors with rsync 3.4.3 when using `--link-dest` over SSH with a relative basis path.

The reproduction stages/generated files into a temporary source directory, syncs them to a remote snapshot directory, then creates a second snapshot using `--link-dest=../snap.1`.

The first baseline snapshot succeeds, but the second snapshot using the relative `--link-dest` basis fails with exit code `23` and many `failed verification -- update discarded` errors.

A later comparison using an absolute `--link-dest` path to an existing snapshot basis succeeds.

This appears possibly related to the recent `failed verification -- update discarded` reports, but this reproduction does not require symlinked receiver directories.

Possibly related issues:

- #715
- #724
- #915

NOTE: This also affects backported rsync versions when they're used on the  Receiver:
```
Debian: 3.4.1+ds1-5+deb13u3 / 3.2.7-1+deb12u5 / 3.2.3-4+deb11u3
Ubuntu: 3.2.7-1ubuntu1.4
```
## Environment

Sender/client:

```text
rsync version 3.4.3 protocol version 32
rsync path: /usr/local/jetapps/usr/bin/rsync
```

Transport:

```text
rsync over SSH
```

Remote receiver:

```text
root@remote-host
```

SSH command used by rsync:

```text
ssh -p 22 -o BatchMode=yes -o StrictHostKeyChecking=accept-new -i /path/to/test_key
```

## Reproduction script

I used the attached script:

[rsync-failed-verification-repro.sh](https://github.com/user-attachments/files/28412015/rsync-failed-verification-repro.sh)

The script does the following:

1. Creates a temporary local staging directory.
2. Generates config-like files into the staging directory.
3. Rsyncs the staged files to a remote `snap.1` directory.
4. Regenerates/copies identical staged files.
5. Rsyncs the staged files to remote `snap.2` using:

```bash
--link-dest=../snap.1
```

6. Continues with additional comparison tests:
   - `snap.5a` using a relative `--link-dest`
   - `snap.5b` using an absolute `--link-dest`

## Command used

```bash
RSYNC_BIN=/usr/local/jetapps/usr/bin/rsync \
[...]

PR #930, opened 2026-05-30
Fix --link-dest=../sibling regression from symlink-race hardening (#915)

### LINKED ISSUE #915: Security fix breaks --link-dest via rsync daemon
I usually perform backups via rsync daemon or to local disks, maintaining a history with hard links.
With the latest security backports in version 3.2.7 (3.2.7-0ubuntu0.22.04.6 on Ubuntu 22.04.5 LTS, 3.2.7-1ubuntu1.4 on Ubuntu 24.04.4 LTS), the --link-dest option does not work in daemon mode:

```
rsync -arAX --delete \
	--files-from="/path/to/files_to_add" -exclude-from="/path/to/exclude_patterns" \
	--link-dest="../01" "/" "rsync://USER@HOST:PORT/DEST/00/"
```

On the deamon host the "01" dir contains the last backup, while the "00" dir is initially empty.
No errors are reported; all files are simply transferred again... and after 2 backup cycles my backup disks are full :-(
In local mode this doesn't happen, and rsync still works correctly:

```
rsync -arAX --delete \
	--files-from="/path/to/files_to_add" -exclude-from="/path/to/exclude_patterns" \
	--link-dest="../01" "/" "/path_to_backup_dir/00/"
```

Downgrading the packages even just on the daemon host (3.2.7-0ubuntu0.22.04.4 on Ubuntu 22.04.5 LTS, 3.2.7-1ubuntu1.2 on Ubuntu 24.04.4 LTS) restores the correct behavior.
I have not had the opportunity to check other versions or systems.

seanedwards:
## Summary

Fixes #915. After the CVE-2026-29518 / CVE-2026-43619 symlink-race hardening, `--link-dest=../01` — the standard rotating-backup layout where the previous backup is a *sibling* of the destination subdir — stopped working with sibling alt-dest dirs. The reported symptom (a `use chroot = no` daemon silently re-transferring every file instead of hard-linking, filling backup disks) turned out to be one facet of a wider regression that also breaks **delta** transfers against such a basis.

## Root cause

The CVE hardening routed the receiver's basis-file open and the generator's alt-dest `stat()`s through `secure_relative_open()`, which confines resolution **beneath the current directory** via `RESOLVE_BENEATH` (openat2 / `O_RESOLVE_BENEATH`) or a per-component `O_NOFOLLOW` walk, rejecting any `..` component.

That confinement is reached at **two** enforcement points:

- The `do_*_at()` wrappers (generator hard-link, in-place output, …) **self-gate**: they fall back to a bare syscall unless `am_daemon && !am_chrooted`, because local/remote-shell transfers have no module boundary and a chroot is its own boundary.
- The receiver's basis-file open (`secure_basis_open` in `receiver.c`) called `secure_relative_open()` **unconditionally**, missing that gate.

In a `use chroot = no` daemon the receiver has `chdir`'d into the transfer destination (a *subdirectory* of the module), so `--link-dest=../01` resolves through `..` to a sibling that is **still inside the module**. The confinement rejected that climb, so:

- the **exact-match hard-link** (via `do_*_at`) found no basis and re-transferred every file → **issue #915**; and
- because of the missed gate, the **delta path** broke more widely: a changed file delta'd against the sibling basis aborts the whole transfer with `got a block match with no basis file` — not only in the no-chroot daemon, but also under `use chroot = yes` (the default; `module_dirlen` is 0 so the path reaches the receiver as a literal `../01`) and in local/remote-shell transfers with `--no-whole-file`.

rsync 3.4.1 (pre-hardening) handled all of these.

## Fix

Both changes narrow the confinement to the one context that needs it — the sanitizing daemon (`am_daemon && !am_chrooted`):

1. **`secure_relative_open()` re-anchors at the module root** — the actual trust boundary — in that context: it prefixes the module-relative path of the current directory and resolves the result beneath `module_dir`. `RESOLVE_BENEATH` then permits a `..` climb that stays inside the module while still rejecting one that escapes it, and any symlink whose target lies outside the module. The current directory is taken from rsync's own `curr_dir[]` (the logical CWD `change_dir()` maintains), **not `getcwd()`**: `curr_dir` is built by the same `normalize_path()`/`clean_fname()` machinery as `module_dir`, so `module_dir` is a guaranteed lexical prefix of it — whereas `getcwd()` returns the physical path with symlinks resolved, so a module whose configured `path` traverses a symlink would make the two disagree and silently skip the re-anchor. Limited to paths containing a `..` component, so forward paths are untouched. On platforms with neither `openat2` nor `O_RESOLVE_BENEATH`, the conservative `..` rejection is retained (link-dest via `..` stays unsupported there, as before).

2. **`secure_basis_open()` gets the same gate the `do_*_at()` wrappers use**: it opens the basis with a bare `do_open()` unless `am_daemon && !am_chrooted`, so the alt-dest resolves against the CWD as a pre-hardening open did. This repairs the chroot-daemon and local `--no-whole-file` delta paths without weakening the no-chroot daemon, which stays on the secure, re-anchored path above.

`t_stub.c` gains weak `curr_dir[]`/`curr_dir_len` definitions so the `tls`/`trimslash` helpers (which link `syscall.o` but not `util1.o`) still link, while helpers that do link `util1.o` use the real symbols.

## Tests

Each was verified to **fail against the unpatched code** and pass with the fix.

- `testsuite/link-dest-daemon_test.py` — no-chroot daemon: the in-module sibling `..` climb must hard-link (inode equality), and a sibling basis that is a *symlink out of the module* must not be followed.
- `testsuite/link-dest-local-delta_test.py` — local `--no-whole-file` delta against a sibling `--link-dest=../01` basis must reconstruct correctly rather than abort (`block match with no basis file`, exit 2, without change 2).
- `testsuite/link-dest-chroot-delta_test.py` — the `use chroot = yes` delta path (the `am_chrooted` branch of change 2), the only test that exercises that branch. It runs as **real root**, or re-exec's under `unshare --user --map-root-user --map-auto` (the range gid map is what lets the daemon's privilege-drop `setgroups()` succeed — plain `--map-root-user` is not enough), and skips where neither is available, matching `testsuite/COVERAGE.md`'s "use chroot = yes … needs root".

The `use chroot = yes` regression was **reproduced directly as root**: against the unpatched receiver the delta transfer aborts (`got a block match with no basis file for "/00/f" (in bak)`, code 2); with the fix it reconstructs the file correctly.

Full `runtests.py` suite: 86 passed, 0 failed (the chroot test runs via `unshare` in the default non-root run).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/receiver.c
+++ b/receiver.c
@@ -99,6 +99,31 @@ static int updating_basis_or_equiv;
  * Anything else is a straight pass-through that preserves the strict contract. */
 static int secure_basis_open(const char *basedir, const char *relpath, int flags, mode_t mode)
 {
+	extern int am_daemon, am_chrooted;
+
+	/* Confinement via secure_relative_open() is only required for the
+	 * sanitizing daemon (am_daemon && !am_chrooted) -- exactly the gate the
+	 * do_*_at() wrappers in syscall.c apply.  In local / remote-shell mode
+	 * there is no module boundary to enforce, and under "use chroot = yes"
+	 * the kernel root is the module, so the chroot is its own boundary
+	 * (CVE-2026-29518 was specific to chroot-disabled daemons).  In both
+	 * cases an alt-dest basis such as --link-dest=../01 must resolve against
+	 * the CWD the way a bare open does (and did before the CVE hardening);
+	 * confining it here would reject the legitimate sibling ".." and strand a
+	 * delta'd file with "block match with no basis file".  Restrict the secure
+	 * path to the one context that needs it. */
+	if (!am_daemon || am_chrooted) {
+		if (basedir) {
+			char fullpath[MAXPATHLEN];
+			if (pathjoin(fullpath, sizeof fullpath, basedir, relpath) >= sizeof fullpath) {
+				errno = ENAMETOOLONG;
+				return -1;
+			}
+			return do_open(fullpath, flags, mode);
+		}
+		return do_open(relpath, flags, mode);
+	}
+
 	if (!basedir && relpath && *relpath == '/') {
 		const char *slash = strrchr(relpath, '/');
 		const char *leaf = slash + 1;
@@ -859,7 +884,7 @@ int recv_files(int f_in, int f_out, char *local_name)
 				basedir = basis_dir[0];
 				fnamecmp = fname;
 				fnamecmp_type = FNAMECMP_BASIS_DIR_LOW;
-				fd1 = secure_relative_open(basedir, fnamecmp, O_RDONLY, 0);
+				fd1 = secure_basis_open(basedir, fnamecmp, O_RDONLY, 0);
 			}
 		}
 
--- a/syscall.c
+++ b/syscall.c
@@ -1782,27 +1782,100 @@ static int secure_relative_open_resolve_beneath(const char *basedir, const char 
 
 int secure_relative_open(const char *basedir, const char *relpath, int flags, mode_t mode)
 {
+	extern int am_daemon, am_chrooted;
+	extern char *module_dir;
+	extern unsigned int module_dirlen;
+	extern char curr_dir[MAXPATHLEN];
+	extern unsigned int curr_dir_len;
+	char modrel_buf[MAXPATHLEN];
+	int reanchored = 0;
+
 	if (!relpath || relpath[0] == '/') {
 		// must be a relative path
 		errno = EINVAL;
 		return -1;
 	}
-	/* Reject any path with a literal ".." component (bare "..",
-	 * "../foo", "foo/..", "foo/../bar", "subdir/.."). The previous
-	 * substring-based check caught only "../" prefix and "/../"
-	 * substring; bare ".." and trailing "/.." escape on the per-
-	 * component walk fallback used by NetBSD/OpenBSD/Solaris/Cygwin
-	 * and pre-5.6 Linux. RESOLVE_BENEATH on Linux/FreeBSD/macOS
-	 * catches some of these in-kernel with EXDEV, but the front
-	 * door must reject them consistently with EINVAL across all
-	 * platforms so callers can rely on the validation. */
-	if (path_has_dotdot_component(relpath)) {
-		errno = EINVAL;
-		return -1;
+
+	/* Sanitizing daemon (am_daemon && !am_chrooted) only: this is the
+	 * sole context in which the confinement here is reached for an
+	 * attacker-influenced path -- the do_*_at() wrappers fall back to a
+	 * bare syscall for the local / remote-shell / chrooted cases, and a
+	 * chroot is its own module boundary.
+	 *
+	 * In this context we have chdir'd into a sub-directory of the module
+	 * (the transfer destination), so the current directory is NOT the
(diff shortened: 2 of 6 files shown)


zv-io:
With all due respect, as I watch this pull request grow in size, I'm trying to read through the code and I'm wondering if you'll be able to deliver a complete and correct fix for the regression reported in #915 if you're using Claude to write and iterate on this?

It would be much more helpful--and instill more confidence--if you were to share a thorough analysis of the original bug and to design a workaround before implementing it. This way, both the code and the regression tests are readable by humans and it reduces the risk of introducing new bugs.

My position as a distro maintainer is more practical: is rsync going to cause us problems or will it work as it has in the past?

Even if this commit bug-free I don't think one could argue that the complexity of this change is acceptable. My question is whether this complexity is necessary.

seanedwards:
@zv-io In full disclosure, I have not contributed to the RSync project before, and I was not deeply familiar with the rsync codebase or development process when I started this. I had the same question you did, if I would be able to deliver a complete and correct fix for the regression using Claude, as without Claude I likely would not have time or attention contribute any code to the project at all. These are the draft results of my work on this issue so far, such as they are.

The PR grew in size as I checked and re-checked any angle I could think of. My original pass only covered cases where rsync is running as a daemon outside of a chroot. In reviewing the work, I identified a common failure mode with coding agents, where they narrowly fix the problem as it was reported, without backing up to look at the bigger picture. So I kept probing, and a later pass revealed a way to reproduce the issue in local mode, so the scope grew to cover that as well.

The total cost of producing this PR was a bit of computer time over my morning coffee. I'd be happy to spend a bit more time on analysis if you could point me to a high-quality example of the kind of document you're looking for. I also won't take any offense if the PR is closed.

zv-io:
> @zv-io In full disclosure, I have not contributed to the RSync project before

I'm not a contributor to this project either but I am grateful for your time in working to address that regression.

> The total cost of producing this PR was a bit of computer time over my morning coffee. I'd be happy to spend a bit more time on analysis if you could point me to a high-quality example of the kind of document you're looking for. I also won't take any offense if the PR is closed.

The related issue #928 may have some additional commentary/insight. What I'd personally like to see is (excuse me if I've missed it) information such as (a) which commit introduced the regression, (b) what is the mechanism by which the regression occurred, (c) a test case demonstrating pre/post that commit, (d) a minimal proposed fix, (e) solicitation of discussion for how to best proceed toward a resolution.

500 lines of force-pushed Claude output is difficult for busy project maintainers to review, and rsync-the-project needs to make it clear that they're interested in preserving the project's credibility right now.

seanedwards:
@zv-io Thanks for the feedback, I appreciate it. I'll put a little bit more time into making this PR legible. I'll leave it open as a draft until I have something more legible to reviewers.

tridge:
@seanedwards many thanks for your efforts on this, I've got a bit different fix pending in #953 - can you please test that and see if it addresses your issue?

tridge:
I've merged #953 to fix this, closing. Please open a new issue if this is not properly fixed

PR #932, opened 2026-05-30
Correct broken test case for decimal numbers in human readable form

michaelmess-de:
Problem:
Decimal numbers in human readable form are subject of the locale settings on the host running the software.
Some locales use "," instead of "." for decimal numbers, breaking that test case.

Solution:
Corrected test case broken for locales that uses , instead of . for decimal numbers in human readable form.
The test now also accepts "," in addition to "." for decimal numbers.


tridge (approved):

tridge:
@michaelmess-de thanks!

PR #933, opened 2026-05-31
Niceness: Reduce the priority of the local and/or remote rsync process by command line option similar to nice or ionice

michaelmess-de:
## Problem:
rsync with plenty of files and data (e. g. for backup purposes) puts load on the client and server machine and people working on these machines might experience a noticeable performance degradation when hardware comes to it's limits. As a workaround the user may renice or ionice the rsync processes to mitigate the issue, but this requires ownership of the running rsync-processes or root access and only works temporarily on the already running process.
Thus in several cases it would be desireable that rsync (e. g. when automatically started by a script or cron job, etc.) gives way for other activities, reducing the impact on those working on the machines.

## Solution:
Added command line options to lower the priority of the rsync process like with renice (lower CPU priority) or ionice (lower IO priority), so that other activities on the machine are not affected by the running rsync job so much.

```
--nice-local             renice local process to low priority
--nice-remote            renice remote process to low priority
--ionice-local           ionice local process to low priority
--ionice-remote          ionice remote process to low priority
--nice, --ionice         set local and remote process to low priority
--no-nice, ...           prepending "no-" turns the given option off
--nice-and-ionice, -Q    same as --nice --ionice
```

The option names are derived from the well known unix tools nice/renice and ionice that can be used to alter priorities of running processes.
With --nice-local or --ionice-local, if there is no support on the local machine OS for reducing priority, despite of a current rsync version, the option is silently ignored.
With --nice-remote or --ionice-remote, if there is no support on the remote machine OS for reducing priority, despite of a current rsync version, the option is silently ignored. If in this case the rsync version on the remote machine does not yet support the new command line options, rsync reports an error.

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -46,7 +46,7 @@ zlib_OBJS=zlib/deflate.o zlib/inffast.o zlib/inflate.o zlib/inftrees.o \
 OBJS1=flist.o rsync.o generator.o receiver.o cleanup.o sender.o exclude.o \
 	util1.o util2.o main.o checksum.o match.o syscall.o log.o backup.o delete.o
 OBJS2=options.o io.o compat.o hlink.o token.o uidlist.o socket.o hashtable.o \
-	usage.o fileio.o batch.o clientname.o chmod.o acls.o xattrs.o
+	usage.o fileio.o batch.o clientname.o chmod.o acls.o xattrs.o niceness.o
 OBJS3=progress.o pipe.o @MD5_ASM@ @ROLL_SIMD@ @ROLL_ASM@
 DAEMON_OBJ = params.o loadparm.o clientserver.o access.o connection.o authenticate.o
 popt_OBJS= popt/popt.o  popt/poptconfig.o \
--- a/configure.ac
+++ b/configure.ac
@@ -13,7 +13,7 @@ AC_CHECK_HEADERS(sys/fcntl.h sys/select.h fcntl.h sys/time.h sys/unistd.h \
     sys/acl.h acl/libacl.h attr/xattr.h sys/xattr.h sys/extattr.h dl.h \
     popt.h popt/popt.h linux/falloc.h netinet/in_systm.h netgroup.h \
     zlib.h xxhash.h openssl/md4.h openssl/md5.h zstd.h lz4.h sys/file.h \
-    bsd/string.h)
+    bsd/string.h sys/resource.h linux/ioprio.h sys/syscall.h)
 AC_CHECK_HEADERS([netinet/ip.h], [], [], [[#include <netinet/in.h>]])
 AC_HEADER_MAJOR_FIXED
 
--- a/configure.ac
+++ b/configure.ac
@@ -1465,6 +1465,8 @@ case "$CC" in
     ;;
 esac
 
+AC_DEFINE_UNQUOTED([COMPILE_TARGET], ["$host"], [String describing the compiled target OS and architecture])
+
 AC_CONFIG_FILES([Makefile lib/dummy zlib/dummy popt/dummy shconfig])
 AC_OUTPUT
 
--- a/main.c
+++ b/main.c
@@ -107,6 +107,8 @@ extern char backup_dir_buf[MAXPATHLEN];
 extern char *basis_dir[MAX_BASIS_DIRS+1];
 extern struct file_list *first_flist;
 extern filter_rule_list daemon_filter_list, implied_filter_list;
+extern int nice_local;
+extern int ionice_local;
 
 uid_t our_uid;
 gid_t our_gid;
@@ -1834,6 +1836,14 @@ int main(int argc,char *argv[])
 	}
 	if (write_batch < 0)
 		dry_run = 1;
+
+	if (nice_local) {
+		renice_me();
+	}
+
+	if (ionice_local) {
+		ionice_me();
+	}
 
 	if (am_server) {
 #ifdef ICONV_CONST
--- a/niceness.c
+++ b/niceness.c
@@ -1,0 +1,48 @@
+/*
+ * Renice or ionice the current process to reduce its impact on the system
+ *
+ * Copyright (C) 2026 Michael Mess
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License along
+ * with this program; if not, visit the http://fsf.org website.
+ */
+
+#include "rsync.h"
+
+void renice_me() 
+{
+#ifdef SUPPORT_RENICE
+	int which = PRIO_PROCESS; // who specifies a Process ID
+	int who = 0; // 0 means current process
+	int prio = 19; // lowest CPU Priority
+	int result = setpriority(which, who, prio);
+	if ( result < 0 ) {
+		// Failed to set priority, TODO: log or inform user, but can be ignored (it's just not so nice).
+	}
+#endif
+}
+
+void ionice_me() 
+{
+#ifdef SUPPORT_IONICE
+	int which = IOPRIO_WHO_PROCESS; // who specifies a Process ID
+	int who = 0; // 0 means current process
+	int class = IOPRIO_CLASS_IDLE;
+	int data = 0; // Ignored when using the IOPRIO_CLASS_IDLE class
+	int ioprio = IOPRIO_PRIO_VALUE(class, data);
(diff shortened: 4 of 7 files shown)


cdlscpmv:
How is it different from prepending `nice` and/or `ionice` commands to `rsync` or `--rsync-path` (in the remote case)?

michaelmess-de:
> How is it different from prepending `nice` and/or `ionice` commands to `rsync` or `--rsync-path` (in the remote case)?

Prepending nice and/or ionice to the remote depends on that these utilities have actually been installed by the user. We do not know, if this is the case on the remote site and if not, the remote command will fail to execute rsync.
But on the other hand, we also do not know, if rsync on the remote site already does know about the option (It might need an upgrade to a current version that supports it). Thus there is currently only little difference. But in the future, when systems have been upgraded, the new options should be known by the remote end, and, if the feature is actually not supported by the remote OS, this would just issue a warning and would not stop us from rsyncing successfully.

With prepending nice/ionice there would be also additional processes to be launched, maybe that could be an issue.
Also when starting nice and ionice remotely, we rely on other software on the remote shells path, which could also be security relevant. And in restricted environments where only rsync is made available as executable, this would definitely fail.
With the feature integrated in rsync we just tell the kernel to reduce the priority directly (if supported on the architecture where the binary was compiled for), this should be safe and efficient.
And in the local case, the user/script has already started rsync, maybe with nice/ionice, maybe not. 
Adding an option to rsync is much easier for the user and thus more reliable and less error-prone than fiddling with nice/ionice on the command line.

cdlscpmv:
> Prepending nice and/or ionice to the remote depends on that these utilities have actually been installed by the user. We do not know, if this is the case on the remote site and if not, the remote command will fail to execute rsync.

`--rsync-path` is a shell command, so you can include tests like `$(which nice >/dev/null 2>/dev/null && echo nice -n 10) rsync` and similarly for `ionice`.

> And in restricted environments where only rsync is made available as executable, this would definitely fail.

The administrator of the restricted environment can be contacted to provide the means to resolve the issue (if they see it as an issue at all, because they may not be concerned about high CPU/disk usage).

michaelmess-de:
Thank you for the workaround with --rsync-path. It is a good workaround for the meantime and works fine with linux.
But most people like it easier.
And maybe the user account could be configured to use another shell, maybe on Windows PowerShell. That could break the --rsync-path workaround or make it more complicated to handle all cases.
Thus it can help some people, but is not a general solution that could be rolled out e.g. in combination with BackupPC or other tools.

devZer0:
thanks for making an enhancement to rsync. 

I'm a little bit surprised, that we have that many new options for adjusting niceness, but without the capability of the original tools, i.e. if you add this complexity, i think adjusting the level to the desired values should also be possible.  

anyhow, i'm not sure if that belongs into rsync, i don't have yet an opinion on that.  
but when it is integrated, i strongly vote for adding levels, too

```
ionice has:

ionice [-c class] [-n level] [

-c, --class class
           Specify the name or number of the scheduling class to use; 0 for none, 1 for realtime, 2 for best-effort, 3 for idle.

-n, --classdata level
           Specify the scheduling class data. This only has an effect if the class accepts an argument. For realtime and best-effort, 0-7 are valid data (priority levels), and 0 represents the highest priority level.
	

nice has:

-n, --adjustment=N

add integer N to the niceness (default 10)
						 
Niceness values range from -20 (most favorable to the process) to 19 (least favorable to the process).

**Comment by michaelmess-de** [comment]
I got the idea, that this might be useful, when I was working on something and felt then the computer's performance went poor after a BackupPC job started. 
As a workaround, I used to type "sudo renice 20 <pid>;sudo ionice -c3 -p<pid>" to get back to performance, but it is always work to do that.
I just wanted to achieve that the rsync process went to the background, not taking any noticeable amount of performance, and I came up with the idea that this would be a very useful feature to be integrated into rsync.
Then I found, this would be much more complex than expected. There are 2 rsync processes, local and remote and there are 2 tools, renice (for CPU usage) and ionice (for disk usage).
The goal was to make it as simple to use as possible, and when analysing the parameter processing I found that I had to implement the no-... options, too, just to be consistent, which resulted in 8 options to switch something on or off.

I thought about implementing levels already, but found the following issues: 
For Renice: Usually we want to lower priority to reduce the performance impact on other activities, increasing priorities require special permissions (root or CAP_SYS_NICE) and would even increase the impact on other activities on the system. Specifying levels would make sense if there were plenty of processes on the system and there is the need to give them different priorities.
If there is a high CPU load, renice would always be beneficial to give way for other activities, the effect would be the biggest, if prio ist lowest.
If there is no CPU load at all, rsync won't suffer from a low prio given.
From my experience I thought, there won't be the need to give special levels to rsync and thus I left that out.

For Ionice: We have four classes: 0 and 2 both are effectively best-effort, 1 is real time and 3 is idle. In addition to that, there are seven priorities that can be set in addition to the class and are only effective for class 1 and 2. For class 0 the priority is taken from the nice value. When we want to bring heavy io load of rsync into the background, usually 3 (idle) is the best option, best effort with other priorities brings only a small, negligible advantage. Running rsync in real-time is possible and needs root, but does it actually make sense?
Experience varies with the different disk types, e. g. a hard disk can be slowed down by random access while an ssd does not have this issue. On a hard disk the idle class really gives an improvement, while best effort won't. Maybe on an ssd experience might be different and might be similar to renice, but I haven't tested that yet.
Thus I have thought, that other values than IDLE won't be useful for ionice and left other values out, just to keep it simple.

But it would be easy for me to change the functionality of the renice switch to contain a renice level instead of just a boolean switch and add another option which takes an argument for the nice priority that is then also used to pass the priority to the remote rsync. The existing options will remain unchanged and renice to 19 when turning nice on. If a level was already set there, it shouldn't be overwritten by turning it on.
For ionice this would be a bit more complicated due to the valid combinations of class and prio, but that should also be possible.

I think it is a good idea for me to develop this on a new branch derived from this, so that the current state can already be reviewed and merged and thus already made available while I am working on the improvements on another branch.

I will add two voting comments (vote by adding a like) for the mentioned improvements.

**Comment by michaelmess-de** [comment]
VOTE: Like to vote for specifying nice levels in rsync

**Comment by michaelmess-de** [comment]
VOTE: Like to vote for specifying IOnice levels in rsync

**Comment by cdlscpmv** [comment]
`ionice` is platform-specific, so adding it to rsync is not a good idea.

**Comment by michaelmess-de** [comment]
> `ionice` is platform-specific, so adding it to rsync is not a good idea.

Yes, ionice is platform-specific, but it wouldn't be the only platform-specific part in rsync.
Other platform-specific parts are for example filesystem specific properties, like path and filenames, hardlinks, symlinks, file ownership and permissions, special files like pipes and sockets that only exist on Unix, timestamps, etc.
Then there are 32bit/64bit, some processor-specific optimisations, network specific switches, etc. 
rsync is using autotools to determine what is supported on a specific platform and preprocessor directives to control which parts can be compiled in.
So I wouldn’t say that including platform-specific things in rsync is not a good idea, it just needs a bit more work to ensure, that the build still works and is consistent, even on platforms that do not support some specific features.

**Comment by michaelmess-de** [comment]
@devZer0 I have now added nice levels in feature-branch https://github.com/RsyncProject/rsync/pull/945

**Comment by tridge** [comment]
@michaelmess-de I think having control of nice and ionice is a good idea, but I'm not keen on all the separate options.
Maybe something like 

```
rsync -av --nice io:7,proc:19,net:7 /home server::homebackup/
```

so combine ionice, process nice and network nice (traffic priority) into one option? Server could refuse with refuse options, server should probably also refuse attempts to raise priority beyond default.

cdlscpmv:
@tridge, the problem with ionice is that it cannot be described with a single number:
1. There are three scheduling classes (realtime, best-effort, idle).
2. Only the best-effort class (used by default) supports numeric priorities but those are supported only by the BFQ scheduler as far as I know. (And BFQ is not the best option for modern SSD devices).

The realtime class requires elevated process privileges, so it's not quite universal.

The idle class can be useful (and it is implemented in this review).

On the client side, rsync can be easily wrapped in `ionice(1)` to adjust its IO priority (with all its nuances).

I am not sure how useful client-specified adjustments would be on the server side though, because performance of the server is a concern of the server and not of the client.

tridge:
> the problem with ionice is that it cannot be described with a single number

it doesn't have to be a single number, I'm just trying to limit the impact of this on the ballooning list of rsync options. Have a look at how --chmod was done, it takes a string as well. 
I'm not really the right person to choose what that string looks like, I'm just the one trying to keep it better contained :-)
also, this needs a testsuite entry (or 2 or 3 ...)

michaelmess-de:
Hi @tridge , hi @devZer0 ,
I have now created https://github.com/RsyncProject/rsync/pull/1011 where everything can be configured using the --nice=... option with parameters. With this approach all nice and all ionice values are now supported. 
However for elevated priorities like negative nice values or ionice realtime elevated permissions are needed (root, CAP_SYS_NICE, CAP_SYS_ADMIN) for the prio change to succeed.
When the request is unsuccessful/unsupported, a warning message is issued and rsync will continue without elevated/changed priorities.

michaelmess-de:
Marked this as draft, it can be closed when https://github.com/RsyncProject/rsync/pull/1011 has been accepted.

michaelmess-de:
Closing this in favour of #1011.

PR #937, opened 2026-05-31
added testing of old static binaries

tridge:
This adds direct testing of old versions of rsync. The existing test suite did backwards compatibility testing by forcing old protocol versions, which is a useful test, but not as complete as actually running old versions.
Adds versions dating back to 2.6, released in 2004


devZer0:
>2.6.0 is the practical floor; older tags need more
>porting to build on a current toolchain.

@tridge , i think that depends, at least i managed to build down to 1.6.5 without too much porting efforts on alpine.  at least i got working binaries.

so if you want to even go down further , here is build instructions for it . unfortunately mail formatting got a little bit broken. can resend a better version if you like. 

https://www.mail-archive.com/rsync@lists.samba.org/msg33961.html

tridge:
> i think that depends, at least i managed to build down to 1.6.5 without too much porting efforts on alpine. at least i got working binaries.

ok, thanks. I think I'm happy with 2.6.0 for now - it didn't find as many issues as I thought it might, but I'm glad to have this as a check for future versions where we may make larger changes

PR #939, opened 2026-06-01
linux: improve openat and openat2 handling

cyphar:
When looking at the usage of `openat` and `openat2` on Linux, I noticed two bugs:

 - `openat(2)` famously does not return `-EINVAL` for unknown flag bits on Linux, so the fallback code to use `O_RESOLVE_BENEATH` if defined would result in all open operations with the "secure resolver" becoming insecure. This would require a fairly convoluted build environment to trigger, so this isn't exploitable (which is why I submitted it here rather than to the security contact email) but if it were triggered it would be a security bug.
 - The usage of `openat2(2)` did not handle `-EAGAIN` at all, which could lead to spurious errors. `openat2(2)` can return `-EAGAIN` when trying to walk into `..` during a scoped lookup like `RESOLVE_BENEATH` and there was a rename or mount operation anywhere else on the system. The reason for this behaviour is a little esoteric but in short I implemented it this way so that userspace could choose their retry policy and (more importantly) the kernel could avoid potentially-unbounded retries that could lead to a DoS. The man page does mention this but it is easy to miss:
   > **EAGAIN** _how.resolve_  contains  either **RESOLVE_IN_ROOT** or **RESOLVE_BENEATH**, and the kernel could not ensure that a ".." component didn't escape (due to a race condition or potential attack).  The caller may choose to retry the **openat2**() call.
   It is possible for some of the concerns raised in other threads about instability with `3.8.3` were caused by this, but nobody has posted `strace` logs so it's hard to know for sure, and I haven't had time to test this myself.

<hr>

I also took a quick look at the `path_has_dotdot_component` usage and I'm not really convinced that it is actually doing anything useful. The comments seem to be Claude-authored so I'm not sure reading them is super helpful, but the upshot of:

```
	/* Reject any path with a literal ".." component (bare "..",
	 * "../foo", "foo/..", "foo/../bar", "subdir/.."). The previous
	 * substring-based check caught only "../" prefix and "/../"
	 * substring; bare ".." and trailing "/.." escape on the per-
	 * component walk fallback used by NetBSD/OpenBSD/Solaris/Cygwin
	 * and pre-5.6 Linux. RESOLVE_BENEATH on Linux/FreeBSD/macOS
	 * catches some of these in-kernel with EXDEV, but the front
	 * door must reject them consistently with EINVAL across all
	 * platforms so callers can rely on the validation. */
```

appears to be that `..` is problematic and `RESOLVE_BENEATH` catches "some of these". This is is incorrect -- `..` cannot escape the starting point with `RESOLVE_BENEATH` (and `O_RESOLVE_BENEATH` as it was modeled after `RESOLVE_BENEATH`), regardless of where you put it in the path. However, you can walk into `..` during a lookup without issue if it remains within the starting root.

Reading the rest of the code, it seems this was actually done because the fallback userspace resolver cannot handle `..` and so the decision was to ban it for everything for "consistency" but this doesn't really make sense -- symlinks can contain `..` so there will always be an inconsistency between the resolver that rejects all symlinks and those that can handle them. It seems like it would be more helpful to most users to permit `..` for `RESOLVE_BENEATH` and only block it in the `O_NOFOLLOW` resolver. And it makes even less sense to restrict `..` for the "yolo" resolver -- it has no protections against symlinks so blocking `..` makes no sense and just causes regressions.

My final patch does this, but I can drop it if you prefer.

For what it's worth, there actually are safe ways of handling `..` in userspace. [libpathrs](https://github.com/cyphar/libpathrs/) (shameless plug!) does this on Linux using `/proc/self/fd/...` to check the logical path against the real path in a very solid way, and [Go's `os.Root`](https://pkg.go.dev/os#Root) has a brute-force approach that in principle works on any platform by restarting the lookup each time they hit `..` and re-applying the logical path walked so far (I plan to adapt the approach for libpathrs at some point so it can be cross-platform as well).

I suspect this will also help alleviate some of the issues people have had with regressions on v3.8.3.

### DIFF
--- a/syscall.c
+++ b/syscall.c
@@ -1677,33 +1677,7 @@ int do_open_nofollow(const char *pathname, int flags)
                                flag name, picked up by the same #ifdef;
                                flag value differs from FreeBSD)
   Other systems fall back to the per-component O_NOFOLLOW walk below.
-
-  The relpath must also not contain any ../ elements in the path.
 */
-
-/* Returns 1 if path has any "/"-separated component that is exactly
- * "..", 0 otherwise. Used by secure_relative_open's front-door
- * validation to reject inputs that the per-component walk fallback
- * would otherwise resolve through ".." -- e.g. bare "..", "foo/..",
- * "subdir/.." -- which RESOLVE_BENEATH-equivalent kernels reject in
- * the kernel but the per-component fallback (NetBSD/OpenBSD/Solaris/
- * Cygwin/pre-5.6 Linux) does not. */
-static int path_has_dotdot_component(const char *path)
-{
-	const char *p = path;
-
-	while (*p) {
-		const char *q;
-		if (*p == '/') { p++; continue; }
-		q = p;
-		while (*q && *q != '/')
-			q++;
-		if (q - p == 2 && p[0] == '.' && p[1] == '.')
-			return 1;
-		p = q;
-	}
-	return 0;
-}
 
 #if defined(__linux__) && defined(HAVE_OPENAT2)
 static int secure_relative_open_linux(const char *basedir, const char *relpath, int flags, mode_t mode)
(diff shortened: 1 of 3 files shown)


tridge:
@cyphar thanks, good to see others looking at this. I also sent you an email if you have a chance to look

tridge:
@cyphar interesting that FreeBSD fails this change with O_RESOLVE_BENEATH. Different semantics? macos does the same.

cyphar:
Yeah the `errno`s are a little different, but I'm a little confused why they appear to be returning `EPERM` -- the FreeBSD man page says that the `errno` should be `ENOTCAPABLE` in this case.

Let me take a look at what they did with `O_RESOLVE_BENEATH`... :/

tridge:
@cyphar can you see if there is anything still needed in this PR? I've merged some basic openat2 fixes including the android hackery

tridge (changes requested):
codex and claude flag errno handling issues in this. Can you look?
```
1. The errno regression is functional, not cosmetic — and there's a second site. I flagged the cleanup: clobber (syscall.c:1876-1881). Codex
  found the same pattern in secure_relative_open_linux() after a failed openat2 + close(dirfd) (syscall.c:1716-1720), and crucially: the outer 
  EAGAIN loop branches on that errno (if (errno==ENOSYS) break; if (errno!=EAGAIN) return -1;). If close() clobbers errno, the retry/fallback 
  decision itself can misfire — a real control-flow bug, not just a wrong error message. → must-fix before merge (save/restore errno at both
  sites).
```

PR #947, opened 2026-06-02
.editorconfig: help maintain consistent style

stsquad:
This is roughly based on QEMU's editorconfig but with different carve outs for the vendored libraries. In practice the popt library looks like it has a fair amount of whitespace damage so I don't know if it would be worth doing a fixup commit to clean it up.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,40 @@
+# EditorConfig is a file format and collection of text editor plugins
+# for maintaining consistent coding styles between different editors
+# and IDEs. Most popular editors support this either natively or via
+# plugin.
+#
+# Check https://editorconfig.org for details.
+#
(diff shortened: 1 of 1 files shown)


wmthornton-dev:
> This is roughly based on QEMU's editorconfig but with different carve outs for the vendored libraries. In practice the popt library looks like it has a fair amount of whitespace damage so I don't know if it would be worth doing a fixup commit to clean it up.

Coming from the RTEMS project, we haven't used .editorconfig files but I can see the usefulness. Might I suggest for the fixup of the popt library that a .clang-format file is generated? Working in RTEMS, I've found that to be extremely helpful for ensuring consistent formatting of files above-and-beyond tabs and spaces. I'm working on a proposed implementation of what that would look like.

PR #952, opened 2026-06-03
fix: Error thrown during dry-run

### LINKED ISSUE #880: --dry-run doesn't work with --mkpath when copying files
When I try dry-running rsync for a single file (gitconfig)  I get this:
```
+ rsync --dry-run --recursive --mkpath --links --perms --times --omit-dir-times /home/peter/src/gitlab/gkwf/home/.config/git/gitconfig /tmp/usr/local/share/git/gitconfig
rsync: [Receiver] change_dir#3 "/tmp/usr/local/share/git" failed: No such file or directory (2)
rsync error: errors selecting input/output files, dirs (code 3) at main.c(829) [Receiver=3.4.1]
+ set +x
```
I know that the canonical answer is that I should do this:
```
+ rsync --dry-run --recursive --mkpath --links --perms --times --omit-dir-times /home/peter/src/gitlab/gkwf/home/.config/git/gitconfig /tmp/usr/local/share/git/
+ set +x
```

But my real goal is to copy the file by another name to the destination:
```
+ rsync --dry-run --recursive --mkpath --links --perms --times --omit-dir-times /home/peter/src/gitlab/gkwf/home/.config/git/gitconfig /tmp/usr/local/share/git/config
rsync: [Receiver] change_dir#3 "/tmp/usr/local/share/git" failed: No such file or directory (2)
rsync error: errors selecting input/output files, dirs (code 3) at main.c(829) [Receiver=3.4.1]
+ set +x
```

The latter nicely succeeds in a live run:
```
+ rsync --recursive --mkpath --links --perms --times --omit-dir-times /home/peter/src/gitlab/gkwf/home/.config/git/gitconfig /tmp/usr/local/share/git/config
+ set +x
$ ls /tmp/usr/local/share/git/
config
```

That is I am unable to verify with a dry run a legitimate rsync operation which eventually, as it turns out, works perfectly. I also checked the destination directory with the help of chattr: rsync does not modify it in the course of the dry runs, therefore I feel it unjust that rsync treats destination files and destination directories differently during dry runs.

  
Environment:
```
$ rpm -q rsync
rsync-3.4.1-5.fc43.x86_64

$ uname -srv
Linux 6.19.13-200.fc43.x86_64 #1 SMP PREEMPT_DYNAMIC Sat Apr 18 20:20:44 UTC 2026
```

userwiths:
closes: #880 

`get_local_name` refers to 2 "modes", when the destination is a folder and when its a file.

It seems that `change_dir` should take into account if we are dry-running in both cases, but it was considering the option only when dest was a folder

### DIFF
--- a/main.c
+++ b/main.c
@@ -832,7 +832,16 @@ static char *get_local_name(struct file_list *flist, char *dest_path)
 		dest_path = "/";
 
 	*cp = '\0';
-	if (!change_dir(dest_path, CD_NORMAL)) {
+	if (dry_run && mkpath_dest_arg && do_stat(dest_path, &st) < 0) {
+		/* --mkpath would have created this parent dir, but a dry run did
+		 * not, so don't chdir into it; flag the destination as not yet
+		 * present (as the dir-creation path above does) so the generator
+		 * doesn't try to compare against the missing tree (#880).  Only
+		 * the missing-parent case is touched, so an ordinary file-to-file
+		 * dry run still itemizes against an existing destination. */
+		dry_run++;
+		change_dir(dest_path, CD_SKIP_CHDIR);
+	} else if (!change_dir(dest_path, CD_NORMAL)) {
 		rsyserr(FERROR, errno, "change_dir#3 %s failed",
 			full_fname(dest_path));
 		exit_cleanup(RERR_FILESELECT);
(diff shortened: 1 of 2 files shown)


steadytao (commented):
Thank you for fixing this. Confirmed fix.

This looks like the right behavioural fix to me. The only thing I would suggest is adding a small regression test for the single-file `--dry-run --mkpath` case so it stays covered.

userwiths:
Not sure how things are handled here, so i used for example what i found in the `testsuite`. Tested with the same file on `master` and on this branch, and as expected it fails on master and passes here.

tridge (approved):

tridge:
@userwiths I did a small bit of rework of your fix, thanks for doing this!

PR #955, opened 2026-06-03
fix: install generated manpages out of tree

### LINKED ISSUE #846: manpages not installed in out-of-tree builds
If I do an out-of-tree build of rsync without md2man working then the manpages don't get installed when build from tarball, which ships pre-generated manpages.

This is because the Makefile's `install` rule looks in the build directory for `rsync.1` but it doesn't exist, it's in the source directory.

steadytao:
Fixes #846 by installing generated manpages from the source directory when they are not present in an out-of-tree build directory.

Release tarballs include pre-generated manpages. With an out-of-tree build configured using `--disable-md2man`, `make install` could install the binaries but skip the manpages because the install rule only checked the build directory.

This keeps the existing build-directory preference then falls back to `$(srcdir)` for `rsync.1`, `rsync-ssl.1`, `rsyncd.conf.5` and the optional `rrsync.1`.

Tested with a v3.4.3 release tarball using:
- `../configure --disable-md2man --with-rrsync`
- `make`
- `make install DESTDIR=...`

Before this change, the install tree missed the generated manpages. After this change, `rsync.1`, `rsync-ssl.1`, `rrsync.1` and `rsyncd.conf.5` are installed.

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -84,12 +84,19 @@ install: all
 	$(INSTALLCMD) -m 755 $(srcdir)/rsync-ssl $(DESTDIR)$(bindir)
 	-$(MKDIR_P) $(DESTDIR)$(mandir)/man1
 	-$(MKDIR_P) $(DESTDIR)$(mandir)/man5
-	if test -f rsync.1; then $(INSTALLMAN) -m 644 rsync.1 $(DESTDIR)$(mandir)/man1; fi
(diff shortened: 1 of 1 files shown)


steadytao:
NetBSD failure looks unrelated to this change. It fails during VM package setup before the rsync build/install path reaches the Makefile change whatsoever.

tridge (approved):
thank you!

PR #957, opened 2026-06-03
receiver: try to chmod the target file when denied opening

shankerwangmiao:
When the target file exists but its permission modes prevent us from opening it for writing, we can try first to chmod it and then open it.

This happens under the following conditions:

1. rsync is started with `--partial` and `--delay-updates`, so that transferred files are saved in the `.~tmp~/` directory
2. the permission modes of some of the files are preventing the owner from writing into it, e.g. 0444
3. the files which has already been transferred is stored in the `.~tmp~/` directory, with their mode set to the same as the source
4. rsync is interrupted, so that such files remain in the `.~tmp~/` directory
5. rsync is started again, while this time, rsync is unable to open the files in the `.~tmp~/` directory for writing, because of the permission modes.

### DIFF
--- a/receiver.c
+++ b/receiver.c
@@ -963,11 +963,40 @@ int recv_files(int f_in, int f_out, char *local_name)
 			if (fd2 == -1 && errno == EACCES) {
 				/* Maybe the error was due to protected_regular setting? */
 				if (use_secure_symlinks)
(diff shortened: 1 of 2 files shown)


tridge:
@shankerwangmiao I made some tweaks to your PR, thanks for your contribution!

PR #966, opened 2026-06-05
receiver: fix NULL deref on the delta discard path

pterror:
`receive_data()` can dereference NULL and crash the receiver on the discard path. When the receiver discards an incoming file's data — `discard_receive_data()` calls `receive_data()` with `fname == NULL`, `fd == -1`, so `mapbuf == NULL` — and the sender sends a block-match token, the `if (!mapbuf)` branch added in `31fbb17d` calls `full_fname(fname)` with `fname == NULL`; `full_fname()` derefs unconditionally → SIGSEGV.

No malice needed — it happens in ordinary operation: any delta transfer where the basis is readable (so the generator sends real checksums and a stock sender emits match tokens) but the receiver can't create its output temp file — a push to a read-only / over-quota / out-of-inodes destination dir, a destination that is a directory where a file is expected, or a `--partial-dir` resume. The receiver child SIGSEGVs; the transfer dies with `code 12`.

Fix: discriminate on `fd`, not `mapbuf`. The error is only meaningful for the real-transfer caller (valid `fd`/`fname`) — emit it (with the now-safe `full_fname(fname)`) only when `fd != -1`; on the discard path (`fd == -1`) absorb the matched bytes (`offset += len; continue;`), restoring pre-`31fbb17d` behavior.

Adds `testsuite/recv-discard-nullderef_test.py` — reproduces it deterministically over a daemon sender/receiver pair (shared-block basis + unwritable destination dir to force the discard path), asserts the receiver doesn't crash; fails (SIGSEGV) on current master, passes with the fix.

Found via LLM-assisted fuzzing; fix and test written with LLM assistance.

### DIFF
--- a/.github/workflows/almalinux-8-build.yml
+++ b/.github/workflows/almalinux-8-build.yml
@@ -62,7 +62,7 @@ jobs:
       # crtimes-not-supported skip matches the other Linux jobs;
       # daemon-chroot-acl and proxy-response-line-too-long skip because
       # the default (secure) transport opens no listening socket.
-      run: RSYNC_EXPECT_SKIPPED=crtimes,daemon-access-ip,daemon-chroot-acl,proxy-response-line-too-long make check
+      run: RSYNC_EXPECT_SKIPPED=crtimes,daemon-access-ip,daemon-chroot-acl,proxy-response-line-too-long,recv-discard-nullderef make check
     - name: check (TCP daemon transport)
       # Second run exercising the real loopback-TCP daemon path.
       run: ./runtests.py --rsync-bin="$PWD/rsync" --use-tcp -j 8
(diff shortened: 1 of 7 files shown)


steadytao (changes requested):
Thank you for the PR.

The fix seems right to me. From the current call graph `fd == -1` is the discard path via `discard_receive_data()` while real output transfers keep the existing protocol inconsistency check.

The remaining issue looks like test/CI shape rather than the receiver fix. Ubuntu, Ubuntu 22.04, AlmaLinux and macOS skip `recv-discard-nullderef` because the jobs run through `sudo` and the chmod precondition cannot force `mkstemp()` failure. Cygwin skips for the same practical reason because the destination directory remains writable despite `0555`.

If the test can force the discard path without relying on DAC permissions, that would be stronger. The PR body mentions the “destination that is a directory where a file is expected” case; that may be a better precondition if it can still produce a match token and reach `discard_receive_data()` under root.

Otherwise, the new skip probably needs to be added to the expected-skip lists for the root/Cygwin jobs, with coverage coming from the runners that can actually exercise the chmod precondition.

Please let me know if you would like to work on this yourself or if you would prefer I resolve these issues.

tridge:
@pterror I've pushed some fixes that hopefully with get CI passing

steadytao (approved):

PR #975, opened 2026-06-06
build: add uninstall targets

### LINKED ISSUE #784: impossible to uninstall rsync with :` sudo make uninstall
[FRA] bonjour ,
 il est impossible de supprimmer rsync ,avec la commane `sudo make uninstall` comme je le montre dans la vidéo , pourriez vous modifier le fichier MakeFile pour que le problème soit corrigé et que tout le monde en profite ?
En attendant je regarde le fichier en question, moi même pour essayer de le modifier sans une intelligence artificielle.
Cordialement
[ENG]
hello,
 it is impossible to delete rsync, with the command `sudo make uninstall` as I show in the video, could you modify the MakeFile file so that the problem is corrected and everyone benefits?
In the meantime, I am looking at the file in question, myself to try to modify it without an artificial intelligence.
Sincerely

steadytao:
Fixes #784 by adding uninstall targets that mirror the files installed by the existing install targets.

This adds:
- `make uninstall` for the main binaries and manpages
- `make uninstall-ssl-daemon` for the stunnel daemon config installed by `install-ssl-daemon`
- `make uninstall-all` for both sets

The target removes known installed files with `rm -f` and does not remove install directories or unrelated files.

### DIFF
--- a/.github/workflows/ubuntu-build.yml
+++ b/.github/workflows/ubuntu-build.yml
@@ -30,6 +30,34 @@ jobs:
       run: ./configure --with-rrsync
     - name: make
       run: make
+    - name: install/uninstall DESTDIR smoke test
+      run: |
+        set -e
+        tmp="$(mktemp -d)"
+        trap 'rm -rf "$tmp"' EXIT
+
+        make install-all DESTDIR="$tmp"
+
+        for path in \
+          /usr/local/bin/rsync \
+          /usr/local/bin/rsync-ssl \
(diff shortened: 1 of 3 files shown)


tridge (commented):
if we don't want this to bitrot I think it needs a CI test

steadytao:
> if we don't want this to bitrot I think it needs a CI test

🤔 Good point. Give me a few minutes.

tridge (approved):

PR #977, opened 2026-06-06
docs: describe files-from comments

### LINKED ISSUE #646: Describe comment syntax for `--files-from` option in the man
At this moment, both `--exclude-from` and `--include-from` options have notions about comments. I mean the phrase

>  Blank lines in the file are ignored, as are whole-line comments that start with ';' or '#'

In fact, `--files-from` uses the same convention for in-file comments, but it is not documented.

Related answer on StackExchange: https://unix.stackexchange.com/a/401447/367446

steadytao:
Fixes #646 by documenting the comment handling for `--files-from`.

The line-based files-from parser ignores blank lines and whole-line comments starting with `;` or `#`. The docs now mention that behaviour and note that it does not apply when `--from0` is used.

### DIFF
--- a/rsync.1.md
+++ b/rsync.1.md
@@ -2390,9 +2390,9 @@ expand it.
 
     The filenames that are read from the FILE are all relative to the source
     dir -- any leading slashes are removed and no ".." references are allowed
-    to go higher than the source dir.  Blank lines are ignored, as are
-    whole-line comments that start with '`;`' or '`#`' (unless [`--from0`](#opt)
-    is specified).  For example, take this command:
+    to go higher than the source dir.  Blank entries are ignored, as are
+    whole-entry comments that start with '`;`' or '`#`'.  For example, take
+    this command:
 
     >     rsync -a --files-from=/tmp/foo /usr remote:/backup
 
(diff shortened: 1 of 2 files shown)


rsync.1.md
```diff
@@ -2390,7 +2390,9 @@ expand it.
 
     The filenames that are read from the FILE are all relative to the source
     dir -- any leading slashes are removed and no ".." references are allowed
-    to go higher than the source dir.  For example, take this command:
+    to go higher than the source dir.  Blank lines are ignored, as are
+    whole-line comments that start with '`;`' or '`#`' (unless [`--from0`](#opt)
```
tridge: can you check the --from0 caveat is correct? maybe a test? claude thinks it is wrong

steadytao: Confirmed from `read_line()`. In line mode, `RL_DUMP_COMMENTS` skips blank lines and whole-line `#` / `;` comments. With `--from0`, `RL_EOL_NULLS` is set and those entries are treated as literal filenames instead. Added a test covering both behaviours.

steadytao: Okay, nevermind my original read was indeed incorrect. Adjusting test behaviours and docs.

tridge (approved):

PR #978, opened 2026-06-06
docs: clarify empty name groupmap matching

### LINKED ISSUE #626: Groupmap doesn't work with empty or wildcard source group
Hi, recently updated from 3.2.3 to 3.2.7 and noticed that groupmap doesn't appear to work when either no source group or a wildcard source group is passed. I believe from the docs that this should work, and it did work in 3.2.3. Also downloaded the 3.3.0 tarball, and the behavior is the same there.

```
jditommasso@jditommasso1:~/scratch/rsync-3.3.0 [last runtime: 0s]
$ ls -l test-src/ test-dest-*
test-dest-explicit:
total 0K

test-dest-wildcard:
total 0K

test-src/:
total 24K
-rw-r--r-- 1 jditommasso users 0 Jul 17 15:00 test
jditommasso@jditommasso1:~/scratch/rsync-3.3.0 [last runtime: 0s]
$ id | awk -F, '{print $1","$2}'
uid=1861(jditommasso) gid=100(users) groups=100(users),20(dialout)
jditommasso@jditommasso1:~/scratch/rsync-3.3.0 [last runtime: 0s]
$ ./rsync --groupmap=users:dialout test-src/test test-dest-explicit/
jditommasso@jditommasso1:~/scratch/rsync-3.3.0 [last runtime: 0s]
$ ./rsync --groupmap=:dialout test-src/test test-dest-wildcard/
jditommasso@jditommasso1:~/scratch/rsync-3.3.0 [last runtime: 0s]
$ ls -l test-src/ test-dest-*
test-dest-explicit:
total 24K
-rw-r--r-- 1 jditommasso dialout 0 Jul 17 15:16 test

test-dest-wildcard:
total 24K
-rw-r--r-- 1 jditommasso users 0 Jul 17 15:16 test

test-src/:
total 24K
-rw-r--r-- 1 jditommasso users 0 Jul 17 15:00 test
jditommasso@jditommasso1:~/scratch/rsync-3.3.0 [last runtime: 0s]
$
```

Let me know if I'm misreading the docs or otherwise doing something wrong here.

steadytao:
Clarifies the `--usermap` and `--groupmap` documentation for empty source-side names. An empty `FROM` value only matches sender-side IDs that have no transmitted name. It is not the wildcard form for named users or groups; `*` should be used for that.

Fixes #626.

### DIFF
--- a/rsync.1.md
+++ b/rsync.1.md
@@ -3016,6 +3016,10 @@ expand it.
 
     >     --usermap=:nobody --groupmap=*:nobody
 
+    An empty **FROM** value matches only sender-side IDs that have no name.  It
+    is not a wildcard for named users or groups; use "`*`" when you want to map
+    every sender-side name.
+
     When the [`--numeric-ids`](#opt) option is used, the sender does not send any
     names, so all the IDs are treated as having an empty name.  This means that
     you will need to specify numeric **FROM** values if you want to map these
--- a/testsuite/ownership-depth_test.py
+++ b/testsuite/ownership-depth_test.py
@@ -11,6 +11,7 @@ we belong to; the uid side then needs root and is skipped.
 # remap only runs as root, so a non-root run exercises the group-only path too.
 fleet_nonroot = True
 
+import grp
 import os
 
 from rsyncfns import (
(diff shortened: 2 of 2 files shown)


tridge:
@steadytao I think we need a test that validates this behavior to make sure the doc is right

steadytao:
> @steadytao I think we need a test that validates this behavior to make sure the doc is right

Can do. Give me a few minutes.

tridge (approved):

PR #980, opened 2026-06-06
Backport regression fixes ready for a 3.4.4 release

tridge:
This is a set of regression fixes from master on top of 3.4.3. The plan is to make a 3.4.4 release with just regression fixes

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -331,6 +331,28 @@ AC_COMPILE_IFELSE([AC_LANG_PROGRAM([[ ]], [[return 0;]])],
 CFLAGS="$OLD_CFLAGS"
 AC_SUBST(NOEXECSTACK)
 
+dnl Only define HAVE_OPENAT2 when both the <linux/openat2.h> header and the
+dnl SYS_openat2 syscall number are present.  syscall.c uses openat2(RESOLVE_BENEATH)
+dnl for the secure resolver on Linux 5.6+; defining it unconditionally broke the
+dnl build on older kernels/headers that lack the header (#924, #905, #900).
+AC_CACHE_CHECK([for openat2],rsync_cv_HAVE_OPENAT2,[
+    AC_COMPILE_IFELSE([
+        AC_LANG_PROGRAM([[
+#include <sys/syscall.h>
+#include <linux/openat2.h>
+]], [[
+struct open_how how;
+how.resolve = RESOLVE_BENEATH;
+return SYS_openat2 + (int)how.resolve;
+]])
+    ],
+    [rsync_cv_HAVE_OPENAT2=yes], [rsync_cv_HAVE_OPENAT2=no])
+])
+if test x"$rsync_cv_HAVE_OPENAT2" = x"yes"; then
+  AC_DEFINE([HAVE_OPENAT2], 1,
+    [Define to use Linux openat2(RESOLVE_BENEATH) in secure_relative_open where available.])
+fi
+
 # arrgh. libc in some old debian version screwed up the largefile
 # stuff, getting byte range locking wrong
 AC_CACHE_CHECK([for broken largefile support],rsync_cv_HAVE_BROKEN_LARGEFILE,[
--- a/configure.ac
+++ b/configure.ac
@@ -888,7 +888,7 @@ AC_FUNC_UTIME_NULL
 AC_FUNC_ALLOCA
 AC_CHECK_FUNCS(waitpid wait4 getcwd chown chmod lchmod mknod mkfifo \
     fchmod fstat ftruncate strchr readlink link utime utimes lutimes strftime \
-    chflags getattrlist mktime innetgr linkat \
+    chflags getattrlist mktime innetgr linkat mknodat mkfifoat \
     memmove lchown vsnprintf snprintf vasprintf asprintf setsid strpbrk \
     strlcat strlcpy stpcpy strtol mallinfo mallinfo2 getgroups setgroups geteuid getegid \
     setlocale setmode open64 lseek64 mkstemp64 mtrace va_copy __va_copy \
--- a/syscall.c
+++ b/syscall.c
@@ -535,7 +535,9 @@ int do_mknod(const char *pathname, mode_t mode, dev_t dev)
 */
 int do_mknod_at(const char *pathname, mode_t mode, dev_t dev)
 {
-#ifdef AT_FDCWD
+	/* HAVE_MKNODAT: older Darwin declares AT_FDCWD but not mknodat(), so
+	 * the at-variant won't build there; fall back to do_mknod() (#896). */
+#if defined(AT_FDCWD) && defined(HAVE_MKNODAT)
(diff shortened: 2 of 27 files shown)


configure.ac
```diff
@@ -331,6 +331,28 @@ AC_COMPILE_IFELSE([AC_LANG_PROGRAM([[ ]], [[return 0;]])],
...
+struct open_how how;
+how.resolve = RESOLVE_BENEATH;
+return SYS_openat2 + (int)how.resolve;
+]])
+    ],
+    [rsync_cv_HAVE_OPENAT2=yes], [rsync_cv_HAVE_OPENAT2=no])
+])
+if test x"$rsync_cv_HAVE_OPENAT2" = x"yes"; then
+  AC_DEFINE([HAVE_OPENAT2], 1,
+    [Define to use Linux openat2(RESOLVE_BENEATH) in secure_relative_open where available.])
```
steadytao, line 353: This needs the generated `configure` updated too. As-is, release builds that run the tracked `./configure` will not detect `HAVE_OPENAT2`, `HAVE_MKNODAT` or `HAVE_MKFIFOAT` so CI is not actually testing the new configure probes.

support/rrsync
```diff
@@ -372,6 +375,7 @@ if __name__ == '__main__':
     only_group.add_argument('-ro', action='store_true', help="Allow only reading from the DIR. Implies -no-del and -no-lock.")
     only_group.add_argument('-wo', action='store_true', help="Allow only writing to the DIR.")
     arg_parser.add_argument('-munge', action='store_true', help="Enable rsync's --munge-links on the server side.")
+    arg_parser.add_argument('-absolute', action='store_true', help="Allow transfer args to use absolute server paths under DIR.")
```
steadytao: I would probably split this out of the 3.4.4 release. `-absolute` is a new restricted-shell option and changes path-validation behaviour so it deserves separate review/tests rather than riding along with the regression fixes.

syscall.c
```diff
@@ -1805,6 +1880,21 @@ int secure_relative_open(const char *basedir, const char *relpath, int flags, mo
 	return secure_relative_open_resolve_beneath(basedir, relpath, flags, mode);
 #endif
 
+	/* Portable fallback only (no kernel RESOLVE_BENEATH): the per-component
+	 * O_NOFOLLOW walk below can't adjudicate ".." safely, so reject it here --
+	 * even for a re-anchored path.  This re-breaks --link-dest=../01 on
+	 * openat2/O_RESOLVE_BENEATH-less platforms (NetBSD/OpenBSD/Solaris/Cygwin/
+	 * pre-5.6 Linux), trading function for safety; on the kernel paths above
+	 * RESOLVE_BENEATH already allowed the in-module climb. */
```
steadytao, line 1888: This limitation should be reflected in the release notes very clearly. The fallback deliberately rejects the relative alt-basis case on platforms without `openat2`/`RESOLVE_BENEATH` so #915 is not fixed uniformly across all 3.4.4 platforms.

steadytao (changes requested):
My main remaining concern is regression coverage. This PR backports a fairly broad set of fixes but no `testsuite/` files changed. I realise the stable branch still uses the old shell suite so the master-side Python tests cannot necessarily be copied directly. Even so, for a release whose purpose is fixing regressions, the higher-risk cases should probably have stable-branch shell-test equivalents.

With the inline issues addressed and some release-branch regression coverage or clear verification notes added this would look much closer to the kind of conservative 3.4.4 release users would expect.

tridge:
@steadytao regarding the testsuite, what I've been doing is using fleettest.py to test this PR branch against the master test suite. That is why I've been merging new options into fleettest to support running the testsuite from one branch against the code from another across the whole fleet of machines.
This is the branch I've created to be the "testsuite branch" for 3.4.x stable:
https://github.com/RsyncProject/rsync/tree/v34-stable-testsuite
I guess I could add a CI test that clones that branch and tests 3.4 properly in CI?
The reason I didn't want to just bring the master testsuite into 3.4 is it is such a large change and I also need to create backport patch sets for 3.2.7 and 3.4.1 for the LTS distro maintainers. 
happy to discuss if you have ideas

configure.ac
thesamesam: The repository doesn't contain generated configure, it's just in dist tarballs? It does have a `configure` that is a stub which runs `configure.sh`, and `configure.sh` isn't checked-in (which is fine).

EDIT: Ah, wait, it fetches some from the website in https://github.com/RsyncProject/rsync/blob/master/prepare-source if missing, rather than running autoreconf or similar always.

tridge: configure is a tiny wrapper around configure.sh. configure is in git, but doesn't change. configure.sh is generated and it not committed.
I know its an odd setup, but this is how its been since 2007 and it does work :-)

syscall.c
tridge: I've added the start of the NEWS updates for 3.4.4 to this branch with a note on this, please check

tridge:
@steadytao I've now added CI jobs for running the master (approximately) test suite on the 3.4 builds on both ubuntu-latest and ubuntu-2204.

t_stub.c
```diff
@@ -39,6 +39,8 @@ int open_noatime = 0;
 size_t max_alloc = 0; /* max_alloc is needed when combined with util2.o */
```
steadytao: This should not stay `0` for the stable release branch. The helper binaries link `util2.o` and `util2.c` treats `max_alloc = 0` as no allocations allowed. Any helper path that reaches `secure_relative_open()`'s per-component fallback can hit `my_strdup()` and abort with `exceeded --max-alloc=0` before testing the intended behaviour.

The separate `v34-stable-testsuite` branch already fixes this by using `(size_t)-1`; I think the PR branch needs the same fix because `make check-progs` builds helpers from this branch, not from the testsuite checkout.

steadytao (changes requested):
Thanks for your work tridge! I rechecked the earlier concerns: `rrsync -absolute` is gone, the #915 platform limitation is now documented in `NEWS.md` and the generated-configure concern does not apply in the same way I thought because `configure` is just a wrapper that regenerates `configure.sh`.

The one remaining blocker I see is `t_stub.c` leaving `max_alloc = 0`. The current stable-suite CI passes on the GitHub Ubuntu kernels because the openat2 path is available but fallback-path helper runs can still trip `my_strdup()` through `util2.o` and abort with `exceeded --max-alloc=0`. Since helper programs are built from this branch, the `(size_t)-1` fix from `v34-stable-testsuite` should be carried here too.

With that fixed, I think this looks perfectly reasonable as our focused 3.4.4 regression-release PR.

configure.ac
steadytao: It is a bit odd but hey it does work 😁

t_stub.c
tridge: thanks! fixed

steadytao (approved):
LGTM!

PR #981, opened 2026-06-06
rsync-ssl: add socat transport

### LINKED ISSUE #869: `rsync-ssl` breaking with openssl 3.6.2
Hello everyone, thanks for maintaining rsync 👋🏻 🥳 

We're using `rsync-ssl` in Arch Linux for mirror synchronization and when upgrading our server fleet we noticed that all synchronizations failed with the following error when upgrading openssl to 3.6.2:

```
Apr 11 13:09:07 mirror.pkgbuild.com systemd[1]: Starting Synchronize repo mirror...
Apr 11 13:09:08 mirror.pkgbuild.com syncrepo[7055]: verify depth is 4
Apr 11 13:09:08 mirror.pkgbuild.com syncrepo[7055]: Connecting to 2a01:4f8:251:598::
Apr 11 13:09:08 mirror.pkgbuild.com syncrepo[7054]: receiving file list ... done
Apr 11 13:09:08 mirror.pkgbuild.com syncrepo[7057]: rsync: connection unexpectedly closed (4154299 bytes received so far) [receiver]
Apr 11 13:09:08 mirror.pkgbuild.com syncrepo[7057]: rsync error: error in rsync protocol data stream (code 12) at io.c(232) [receiver=3.4.1]
Apr 11 13:09:08 mirror.pkgbuild.com syncrepo[7054]: rsync: [generator] write error: Broken pipe (32)
Apr 11 13:09:08 mirror.pkgbuild.com syncrepo[7054]: rsync error: error in socket IO (code 10) at io.c(849) [generator=3.4.1]
Apr 11 13:09:08 mirror.pkgbuild.com systemd[1]: syncrepo.service: Main process exited, code=exited, status=10/n/a
Apr 11 13:09:08 mirror.pkgbuild.com systemd[1]: syncrepo.service: Failed with result 'exit-code'.
Apr 11 13:09:08 mirror.pkgbuild.com systemd[1]: Failed to start Synchronize repo mirror.
```

When downgrading back to 3.6.1 everything worked again. Any idea how we could start to debug this in more detail? Adding any verbosity flags didn't show more useful information. 🤔 

The script that we're using `rsync` with is the following:
```bash
#!/bin/bash

target="/srv/ftp"
lock="/run/lock/syncrepo.lck"
source_url='rsync://rsync.archlinux.org/ftp_tier1'
lastupdate_url='https://rsync.archlinux.org/lastupdate'

[ ! -d "${target}" ] && mkdir -p "${target}"

exec 9>"${lock}"
flock -n 9 || exit

rsync_cmd() {
        local -a cmd=(rsync-ssl --type=openssl --verbose -rlptH --safe-links --delete-delay --delay-updates
                "--timeout=600" --no-motd)

        if stty &>/dev/null; then
                cmd+=(-h -v --progress)
        else
                cmd+=("--info=name1")
        fi

        "${cmd[@]}" "$@"
}

# if we are called without a tty (cronjob) only run when there are changes
if ! tty -s && [[ -f "$target/lastupdate" ]] && diff -b <(curl -Ls "$lastupdate_url") "$target/lastupdate" >/dev/null; then
        # keep lastsync file in sync for statistics generated by the Arch Linux website
        rsync_cmd "$source_url/lastsync" "$target/lastsync"
        exit 0
fi

rsync_cmd \
        --exclude="/other" \
        --exclude="/sources" \
        --exclude="*-debug/" \
        "${source_url}" "${target}"
```

If you're interested, here is the incident in detail: https://gitlab.archlinux.org/archlinux/infrastructure/-/issues/818

cc @klausenbusk @eworm-de  @jelle

steadytao:
Fixes #869 by adding `socat` as an `rsync-ssl` transport.

`rsync-ssl` currently supports `openssl s_client`, `gnutls-cli` and `stunnel`. The failure reported in #869 points at the `openssl s_client` path breaking as a transport with OpenSSL 3.6.2 while the investigation in the issue found that `stunnel` and `socat` avoid the failure mode. The linked OpenSSL issue, openssl/openssl#30925 still appears to be the upstream place for the `s_client` regression so this keeps the existing OpenSSL path for compatibility rather than trying to work around that buffering behaviour inside rsync.

This adds `--type=socat` and `RSYNC_SSL_SOCAT`.

The socat transport uses:
- `OPENSSL:host:port`
- `commonname=host`
- `snihost=host` for host names
- `no-sni=1` for IP literals
- `verify=1` when `RSYNC_SSL_CA_CERT` is unset
- `verify=0` when `RSYNC_SSL_CA_CERT` is set to an empty value
- `cafile=...` when `RSYNC_SSL_CA_CERT` is set to a path
- `cert=...` and `key=...` for client certificate configuration

The automatic helper search still prefers `openssl` first, preserving the existing default for users who have OpenSSL installed. If OpenSSL is unavailable, it now tries `socat` before `stunnel`. That fallback order is deliberate; in this script, stunnel cannot use the default CA certificate collection while socat can still perform verification without requiring an explicit CA file.

The new testsuite coverage uses fake helper binaries to verify the wrapper/helper contract without requiring a live TLS daemon. It covers explicit `--type=socat`, socat command construction, SNI handling, CA/cert/key options, disabled verification, OpenSSL-first default selection and socat fallback when OpenSSL is absent.

### DIFF
--- a/rsync-ssl
+++ b/rsync-ssl
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 
-# This script uses openssl, gnutls, or stunnel to secure an rsync daemon connection.
+# This script uses openssl, gnutls, socat, or stunnel to secure an rsync daemon connection.
 
 # By default this script takes rsync args and hands them off to the actual
 # rsync command with an --rsh option that makes it open an SSL connection to an
@@ -31,13 +31,16 @@ function rsync_ssl_run {
 
 function rsync_ssl_helper {
     if [[ -z "$RSYNC_SSL_TYPE" ]]; then
-	found=`path_search openssl stunnel4 stunnel` || exit 1
+	found=$(path_search openssl socat stunnel4 stunnel) || exit 1
 	if [[ "$found" == */openssl ]]; then
 	    RSYNC_SSL_TYPE=openssl
 	    RSYNC_SSL_OPENSSL="$found"
 	elif [[ "$found" == */gnutls-cli ]]; then
 	    RSYNC_SSL_TYPE=gnutls
 	    RSYNC_SSL_GNUTLS="$found"
+	elif [[ "$found" == */socat ]]; then
+	    RSYNC_SSL_TYPE=socat
+	    RSYNC_SSL_SOCAT="$found"
 	else
 	    RSYNC_SSL_TYPE=stunnel
 	    RSYNC_SSL_STUNNEL="$found"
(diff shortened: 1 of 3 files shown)


steadytao:
Overlooked BSD. Pushing a fix shortly for CI fails.

PR #982, opened 2026-06-06
:sparkles: Add pre-commit workflow for code quality checks

mdeweerd:
- Add pre-commit workflow for code quality checks
- Add pre-commit configuration file
- Add cache directory for cppcheck
- Add pre-commit hooks for code formatting, linting, and static analysis


mdeweerd:
This adds several static code checks, file formatting tools, etc.
Some are disabled.

I've committed some of the automatic formatting changes and spelling corrections.  In the spelling corrections, the conversion from "uptodate" to "up-do-date" is one that kind of has a functional impact.

With regards to cpplint and cppcheck, more notices can be disabled, or some could be fixed.  Again, this depends on preference.

I actually had found that a processor instruction itself did not modify the register for a 32-bit shift.

Let me know if & how you want to introduce this.

steadytao:
Thanks for doing this. I like the idea behind the goal but I would not merge this as-is.

This is a **very** large and slow pre-commit setup and it mixes tooling, CI, spelling fixes, executable-bit changes, shell formatting, C formatting and generated/configure-style file churn in one PR. That makes the useful static-analysis signal hard to review and risks making local commits expensive for contributors.

I would split this **a lot**; cheap optional pre-commit hooks locally, heavier cppcheck/cpplint checks in CI and separate PRs for spelling, executable bits and formatting churn.

mdeweerd:
The main goal of the current PR is to show the global idea - different authors have different preferences, so this shows the big picture.

I've added two PRs for the moment - one for the executable bits, another for the whitespace.

The codespell fix can be conflicting if done separately, so I'll update that after the merge of these extra PRs.
Same for the shell script beautyfication.

I'll set cppcheck and cpplint to manual.  cpplint does not actually require a lot of time to run and does not need external too installation as cppcheck does.

mdeweerd:
+ At this time I did not include python formatting or static analysis hooks to the pre-commit configuration as I started of a "template" for c-projects.

PR #1001, opened 2026-06-12
Initial LTFS order awarness functionality

hugo-hur:
## Motivation

I'm running several LTFS formatted LTO tapes for incremental backups. Backupping to tape has been working fine with rsync (using the --whole-file and -t options) but reading from the tape is basically impossible due to the reasons stated below.

## Summary

Adds a `--ltfs` option that makes rsync read an LTFS-tape source in **physical
block order** instead of name order, so a restore streams the tape forward in a
single pass instead of seeking back and forth ("shoe-shining").

On an LTFS volume, each file's metadata — name, size, mtime, and the block where
its data begins — lives in the volume index and is cheap to read. File *content*,
however, requires physically positioning the tape. rsync's normal name-sorted
traversal bears no relation to physical layout, so an unordered restore can take
many times longer than one forward pass.

## What it does

- The **sender** reads each file's starting block from the `ltfs.startblock`
  virtual xattr (also honoring a `user.ltfs.startblock` alias) and transmits it
  as a new optional 64-bit file-list extra.
- The **generator** drives the transfer in ascending start-block order. Entries
  with no start block (directories, symlinks) sort first, which front-loads
  creation of the destination directory tree before the bulk data read.

## Behavior & safeguards

`--ltfs` is opinionated about the things that would silently defeat it:

- **Implies `--whole-file`** — a delta transfer would re-read the source off
  tape anyway.
- **Forces `--no-inc-recursive`** — the complete file list is needed before a
  read order can be chosen.
- **Refuses `--checksum`** — it would read every byte of every file off the tape
  just to decide what to transfer.
- **Implies `--times`** (just like `--archive`) — the whole benefit is that the index
  serves size+mtime for free so unchanged files are skipped without reading
  content; that only holds across runs if mtimes are preserved. A later
  `--no-times` overrides it but emits a warning.

## Compatibility / requirements

- Reading the start block needs xattr support, so the feature is gated on
  `SUPPORT_XATTRS`: the file-list extra is only registered when xattrs are
  available, and a build without them **refuses `--ltfs`** (like `--crtimes`)
  rather than silently no-op'ing.
- Over ssh, both ends must have this functionality (sender reads the start block, generator
  orders the reads); `--ltfs` is forwarded to the server side.
- Only the **read (restore)** direction is affected.

## Testing

- New `testsuite/ltfs_test.py` exercises ordering on an ordinary filesystem via
  the `user.ltfs.startblock` alias: round-trip integrity, physical-block read
  order across subdirectories (directories handled first), the implied `-t` /
  `--no-times` warning, and that `--checksum` is refused. It skips cleanly when
  the build lacks xattr support or the platform/filesystem can't set a `user.*`
  xattr (e.g. Windows, where `os.setxattr` is absent).
- Full suite green; builds warning-free both with and without
  `--disable-xattr-support`.
- **Validated against a real LTFS volume:** `--ltfs` read three files
  whose start blocks ran opposite to name order in correct forward order
  (verified via `strace` on the drive), all MD5s matched the tape, and a second
  `-t` pass skipped every file from index metadata alone — zero content reads.

## Not in scope (possible follow-ups)

- Add partition check to sorting for special cases where small files are persisted in the LTFS index partition instead of data partition.


rsync.1.md
```diff
@@ -915,6 +916,37 @@ expand it.
...
+    ascending block order, so the drive makes one forward pass.  Files whose
+    start block is unknown (directories, symlinks, anything not on tape) are
+    handled first, which conveniently creates the destination directory tree
+    before the bulk data read begins.
+
+    Because the whole point is to avoid re-reading tape data, this option
+    implies [`--whole-file`](#opt) (a delta transfer would re-read the source
+    file anyway) and forces [`--no-inc-recursive`](#opt) (the complete file
+    list is needed before the read order can be chosen).  It also refuses
+    [`--checksum`](#opt), which would read every byte of every file off the
```
varesa, line 940: Something I commented offline but I'll mention here as well:

I would not refuse `--checksum` since there are use cases where one would like to ensure that the contents of e.g. an offsite mirror match what is on tape. The tape drive might be even two orders of magnitude faster than the internet link to an offsite archive it makes sense to 
a) ensure matching checksums 
b) minimize amount of data transferred over the internet link

an extra linear scan on a fast drive is a lot smaller issue than shoeshining

hugo-hur: The `--checksum` functionality could be made to check the `ltfs.hash.*` or `user.ltfs.hash.*`  values against the receiver provided ones when used same time with the `--ltfs` flag. These are free to read as they live in the LTFS index which resides in memory when the cartridge is mounted.

Caveat is that hashing is not enforced by the filesystem and the writing application is responsible for storing those hash values correctly and there is no checking of the actual stored hash value against the data in the format specification or the reference implementation of the filesystem. However I would think it is safe to rely on that hash xattr data if it exists for a file (you can blame the software that wrote it if it does not match). If any of the xattr hashes for the formats rsync requires do not exist I would suggest a few options how the rsync would proceed in the case the `--checksum` flag is used:

- Read the files for which the hashes do not exist in tape block order and calculate them → then store to the file xattrs if the filesystem is mounted rw and compare them on the fly against the receiver provided ones
- Fall back silently to size+mtime quick-check for files missing a usable hash xattr treating them as "transfer if mtime or size differs", which errs on the side of correctness and avoids any tape seek for the comparison pass itself
- Refuse the run with a clear error when --checksum --ltfs is used and the tape is not mounted rw and some files lack hash xattrs so the user knows they cannot populate the missing values and the run would degrade.

Also there is a problem with compatible hash algorithms: Rsync negotiates the strongest mutually supported checksum algorithm between sender and receiver. The LTFS format specification defines only five hash types for on-tape storage: ltfs.hash.crc32sum, ltfs.hash.md5sum, ltfs.hash.sha1sum, ltfs.hash.sha256sum, ltfs.hash.sha512sum. xxHash (which rsync likes to use) is not in the LTFS spec and no archiver should write it there. As such, the intersection between what rsync negotiates and what the LTFS index carries is therefore: md5 (always available), and sha1 (if rsync is built against OpenSSL). In practice most backup software writes ltfs.hash.md5sum and increasingly ltfs.hash.sha256sum; better algorithms than sha1 are not available for rsync currently which is suboptimal.

Suggestions are welcome.

varesa: As a user I'd expect --checksum to actually read the data instead of relying on a stored checksum which could be affected by a number of factors including software bugs in the writer.

I'd prefer if such an optimization would be opt-in

hugo-hur: Okay I think this would be a subject of a separate pr in the future.

ltfs.c
```diff
@@ -0,0 +1,136 @@
...
+#endif
+
+extern int ltfs_mode;
+extern int startblock_ndx;
+
+/* Return the LTFS starting block of fname, or -1 if it cannot be determined.
+ * The attribute value is an ASCII decimal block number. */
+int64 ltfs_startblock(const char *fname)
+{
+#ifdef SUPPORT_XATTRS
```
tridge, line 54: I think we need a SUPPORT_LTFS that depends on SUPPORT_XATTRS and makes the option go away completely in the build if not enabled.

hugo-hur: Okay lets see what I can do about that. Some support is needed also in systems without the actual xattrs support in order to be able to pass the flag to possible remote side though.

hugo-hur: Added couple of test for the ltfs specific changes and fixed Cygwin issue, tested the cygwin build myself after reconfigure. Cannot say about the mac version because I don't have access to mac, but I assume these same changes should fix it too. Also added separate test and ltfs feature enablement based on xattrs enablement. Also added separate disablement flag in the autotools chain that can be passed to configure.

hugo-hur:
Rebased on top of changes on the master branch

mosesn292-hue:
hugo-hur:ltfs-aware

hugo-hur:
Changed to draft for now as I found out some problems with testing I have to look at and test other things too locally (mainly compatibilities between rsync versions).

mosesn292-hue:
BTC from $200 to get $250 1 hour ,,, bc1qcg7k59vtv6xd2przmcrynvy6s2qkwa2g0t3jya

hugo-hur:
Now tests pass on Debian and Cygwin, should pass also elsewhere (famous last words). The `--disable-ltfs` flag also tested working for the configure. Added more test cases that are automatically skipped if built without support for this new feature.

hugo-hur:
Also worth noting that currently Cygwin configure flags xattrs as linux style available. I'm not sure if that works as expected...

hugo-hur:
Rebased to master once more. Also added another pr https://github.com/RsyncProject/rsync/pull/1007 for sha256 support for OpenSSL builds for possibly helping on the future `--checksum`  functionalities.

steadytao:
Should likely require the complete xattr value to be a valid non-negative decimal and reject ERANGE or trailing characters. As written, `123junk` and an overflowing value can be accepted as physical positions and silently produce incorrect tape order.

PR #1007, opened 2026-06-14
Added the more modern sha256 option for the --checksum algorithms.

hugo-hur:
Added modern algorithm for the checksum. Works out of the box with builds that have OpenSSL libs enabled.

### DIFF
--- a/checksum.c
+++ b/checksum.c
@@ -60,6 +60,9 @@ struct name_num_item valid_checksums_items[] = {
 #ifdef SHA_DIGEST_LENGTH
 	{ CSUM_SHA1, NNI_EVP, "sha1", NULL },
 #endif
+#ifdef SHA256_DIGEST_LENGTH
+	{ CSUM_SHA256, NNI_EVP, "sha256", NULL },
+#endif
 	{ CSUM_NONE, 0, "none", NULL },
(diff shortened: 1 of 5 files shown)


steadytao (changes requested):
LGTM but docs would need to be updated too; the `--checksum-choice` section still lists the available checksum names without sha256. Would need to remember a NEWS update once we are ready as well.

hugo-hur:
Added to the list of algorithms and something to the NEWS.md (feel free to edit or leave the NEWS.md unmerged).

steadytao:
NEWS would be updated pre-release nonetheless, thank you for the PR.

steadytao (approved):
LGTM. @tridge

hugo-hur:
@tridge can this be merged?

steadytao:
Looking back at this one. Could you rebase and add focused coverage for a SHA-256 transfer, checksum negotiation with an older peer and the compiled algorithm list when OpenSSL SHA-256 support is absent?

steadytao (approved):

steadytao:
Ah. Implementation is fine. Ubuntu failure just needs protocol 29 to forcibly assert itself so there isn't an issue when negotiating md4/md5.

PR #1013, opened 2026-06-19
Fix Solaris xattr write loop after partial write

ThatByteGuy:
This fixes a bug in the Solaris xattr backend's retry loop.

In `lib/sysxattrs.c`, the loop advances `bufpos` after a successful partial write, but subsequent retries continue to pass the original `size` value to `write()` rather than the remaining byte count.

As a result, if a short write occurs, the next iteration can attempt to write more bytes than remain in the source buffer.

The fix changes:

```c
write(attrfd, (char*)value + bufpos, size);
```

to:

```c
write(attrfd, (char*)value + bufpos, size - bufpos);
```

This matches the corresponding logic in `read_xattr()` and the existing retry-loop pattern used elsewhere in the file.

Notes:

* Solaris-specific code path
* No behavioral changes on successful full writes
* Build and test results unchanged after the fix
* The issue appears to date back to the initial Solaris xattr implementation

### DIFF
--- a/lib/sysxattrs.c
+++ b/lib/sysxattrs.c
@@ -227,7 +227,7 @@ int sys_lsetxattr(const char *path, const char *name, const void *value, size_t 
 		return -1;
 
 	for (bufpos = 0; bufpos < size; ) {
-		ssize_t cnt = write(attrfd, (char*)value + bufpos, size);
+		ssize_t cnt = write(attrfd, (char*)value + bufpos, size - bufpos);
 		if (cnt <= 0) {
 			if (cnt < 0 && errno == EINTR)
 				continue;


steadytao (changes requested):
The `size - bufpos` fix seems right for the short-write case so I am happy with that but one nearby issue I noticed while reading the loop is that `bufpos` is a `size_t` so the existing `bufpos = -1` failure sentinel wraps to `SIZE_MAX` and the final `bufpos > 0 ? 0 : -1` can report success after a write failure. 

Not caused by this PR but probably worth fixing in the same area as you are already touching the loop 🙂

ThatByteGuy:
Thanks for the review.

I've fixed the issue you identified and verified the change locally. Would you like me to add the update to this PR, or is there another preferred way to submit the additional commit?

(Here is the commit link: https://github.com/AlphaGlider25/rsync/commit/2a8bbb7792bf741b6c3c08041e33e9fccfffdc7a)

steadytao (approved):
LGTM.

steadytao:
Thank you, I will have a look at 1014 now!

PR #1014, opened 2026-06-19
Fixed as Per Request in prior PR https://github.com/RsyncProject/rsync/pull/1013

ThatByteGuy:
Update:

During review it was pointed out that the Solaris write loop also uses
`bufpos = -1` as an error sentinel even though `bufpos` is a `size_t`.

This wraps to `SIZE_MAX`, allowing the final return expression to
report success after a failed write.

The updated patch returns `-1` directly on non-retryable write errors
instead of using a sentinel value.

### DIFF
--- a/lib/sysxattrs.c
+++ b/lib/sysxattrs.c
@@ -227,19 +227,19 @@ int sys_lsetxattr(const char *path, const char *name, const void *value, size_t 
 		return -1;
 
 	for (bufpos = 0; bufpos < size; ) {
-		ssize_t cnt = write(attrfd, (char*)value + bufpos, size);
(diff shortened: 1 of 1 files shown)


steadytao (approved):
Much appreciated. LGTM.

ThatByteGuy:
is it a possibility for this pr to be merged to rsync? or is there a process i may not be aware of

steadytao:
> is it a possibility for this pr to be merged to rsync? or is there a process i may not be aware of

We are currently busy with the next release then we will begin looking at external contributions more closely.

@tridge Adding another docket 😅

ThatByteGuy:
is it a possibility to join the project?

tridge:
I've fixed the commits (single commit is correct)

Issue #1025, opened 2026-07-12
Regression: 3.4.4 fails to build on macOS 10.6: `syscall.c: error: too few arguments to function ‘do_lchown’`


barracuda156:
```
:notice:configure --->  Configuring rsync
:debug:configure gcc_compilers.tcl not found in ports tree, using built-in selections
:debug:configure Preferred compilers: gcc-4.2 llvm-gcc-4.2 gcc-4.0 macports-gcc-16 macports-gcc-14 macports-gcc-13 macports-gcc-12 macports-gcc-11 macports-gcc-10 macports-gcc-9 macports-gcc-8 macports-gcc-7 macports-gcc-6 macports-gcc-5
:debug:configure Using compiler 'Xcode GCC 4.2'
:debug:configure elevating privileges for configure ccache: euid changed to 0, egid changed to 0.
:debug:configure dropping privileges: euid changed to 503, egid changed to 501.
:warn:configure ccache_dir /opt/local/var/macports/build/.ccache could not be initialized; disabling ccache: couldn't execute "ccache": no such file or directory
:debug:configure Executing proc-pre-org.macports.configure-configure-0
:debug:configure system -W /opt/local/var/macports/build/rsync-62cd6a74/work/rsync-3.4.4: ./prepare-source
:info:configure make: Nothing to be done for `conf'.
:debug:configure Executing org.macports.configure (rsync)
:debug:configure Environment: 
:debug:configure CC='/usr/bin/gcc-4.2'
:debug:configure CC_PRINT_OPTIONS='YES'
:debug:configure CC_PRINT_OPTIONS_FILE='/opt/local/var/macports/build/rsync-62cd6a74/work/.CC_PRINT_OPTIONS'
:debug:configure CFLAGS='-pipe -Os -arch ppc'
:debug:configure CPATH='/opt/local/include'
:debug:configure CPPFLAGS='-I/opt/local/include'
:debug:configure CXX='/usr/bin/g++-4.2'
:debug:configure CXXFLAGS='-pipe -Os -arch ppc'
:debug:configure DEVELOPER_DIR='/Developer'
:debug:configure F90FLAGS='-pipe -Os -m32'
:debug:configure FCFLAGS='-pipe -Os -m32'
:debug:configure FFLAGS='-pipe -Os -m32'
:debug:configure INSTALL='/usr/bin/install -c'
:debug:configure LDFLAGS='-L/opt/local/lib -Wl,-headerpad_max_install_names -arch ppc'
[...]

PR #1026, opened 2026-07-12
syscall: build without AT_SYMLINK_NOFOLLOW

steadytao:
Addresses the upstream portion of #1025.

`do_lstat_at()` selected its `fstatat()` implementation whenever symlink support was enabled. Older systems such as macOS 10.6 support symlinks but do not define `AT_FDCWD` or `AT_SYMLINK_NOFOLLOW` so the unconditional reference to `AT_SYMLINK_NOFOLLOW` prevented `syscall.c` from compiling. This falls back to `do_lstat()` unless both `AT_FDCWD` and `AT_SYMLINK_NOFOLLOW` are available. The same availability check is applied to `do_lchown_at()` because its `fchownat()` implementation also requires `AT_SYMLINK_NOFOLLOW`.

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -65,12 +65,18 @@ CHECK_SYMLINKS = testsuite/chown-fake_test.py testsuite/devices-fake_test.py \
 
 # Objects for CHECK_PROGS to clean
 CHECK_OBJS=tls.o testrun.o getgroups.o getfsdev.o t_stub.o t_unsafe.o t_chmod_secure.o t_secure_relpath.o trimslash.o wildtest.o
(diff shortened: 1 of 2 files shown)


steadytao:
@tridge Should certainly be in 3.5

tridge:
@steadytao I don't think this will fix all the issues with old macos. I've dug out an old mac-mini I have here and I'm working on getting a 10.6 based build going on it.

tridge (approved):
turns out it does build with 10.6, merging for 3.5.0, thanks!

PR #1029, opened 2026-07-19
util: restore ".." collapsing in clean_fname()

dr-who:
## Problem

Commit 9e089846 ("util: fixed issue in clean_fname()") fixed a genuine read-before-buffer in the `".."` back-up loop, but the replacement loop stops in a **different place** than the one it replaced, and the surrounding test/assignment were carried over unchanged.

The old loop pre-decremented, so it stopped with `s` pointing **at** the preceding `/`:

```c
while (s > limit && *--s != '/') {}
```

The new loop stops with `s` pointing at the **start of the preceding component**, one byte past that `/`:

```c
while (s > limit && s[-1] != '/')
	s--;
```

But the follow-up is unchanged, so it is now off by one:

```c
if (s != t - 1 && (s <= name || *s == '/')) {
	t = (s == name) ? name : s + 1;
```

`*s` is a component character and is essentially never `/`, so the collapse now only happens via the `s == name` case — a single leading component with no slash before it. **Every other form silently stopped collapsing:**

| input | current | expected |
|---|---|---|
| `a/../b` | `b` | `b` (still worked) |
| `/a/../b` | `/a/../b` | `/b` |
| `a/b/../c` | `a/b/../c` | `a/c` |
| `/usr/local/../bin` | `/usr/local/../bin` | `/usr/bin` |
| `x/y/../../z` | `x/y/../../z` | `z` |
| `one/two/three/../../four` | `one/two/three/../../four` | `one/four` |

## Impact

`clean_fname(..., CFN_COLLAPSE_DOT_DOT_DIRS)` is rsync's canonical path normalizer — used for filter and merge-file names and for the `--backup-dir`, `--partial-dir` and `--temp-dir` options — so any such path containing an interior `..` is left un-normalized.

User-visible example (a merge filter that can only be opened if `..` is collapsed):

```
$ rsync -a --filter='merge a/none/../c' src/ dst/
rsync: [client] failed to open exclude file a/none/../c: No such file or directory (2)
rsync error: error in file IO (code 11) at exclude.c(1482)
```

Before 9e089846 this collapsed to `a/c` and worked.

## Fix

Adjust the two spots for where the new loop actually stops: test the byte **before** `s` for the `/` (or that we reached the start of the name), and set `t = s`.

This **keeps the underflow fix** — `s[-1]` is only read when `s` is past `name`, and the `s == name` test short-circuits — while restoring the long-standing behaviour.

The `CFN_REFUSE_DOT_DOT_DIRS` path returns before this code and is unaffected, as is `sanitize_path()` (the separate daemon-side `..` handler), so this is a path-normalization/file-selection fix, not a security change.

## Test

`testsuite/clean-fname-collapse_test.py` drives the collapse through a merge-filter filename: the filter file lives at `a/c` but is named `a/none/../c` with **no `a/none` directory**, so it can only be opened if `..` is collapsed. It also asserts the filter was actually applied. It **fails before this change and passes after**; full testsuite otherwise unchanged (106 passed, 0 failed).

### DIFF
--- a/testsuite/clean-fname-collapse_test.py
+++ b/testsuite/clean-fname-collapse_test.py
@@ -1,0 +1,68 @@
+#!/usr/bin/env python3
+"""Regression test for clean_fname()'s ".." collapsing.
+
+clean_fname(name, CFN_COLLAPSE_DOT_DOT_DIRS) is rsync's canonical path
+normalizer.  Commit 9e089846 ("util: fixed issue in clean_fname()") fixed a
+read-before-buffer in the ".." back-up loop, but the new loop leaves 's'
+pointing at the *start* of the preceding component (just past its '/'),
+whereas the old loop left it *at* the '/'.  The surrounding test and
+assignment (`*s == '/'` / `t = s + 1`) were not adjusted for that one-byte
+shift, so the collapse silently stopped working for every path except a
+single leading no-slash component:
+
+    a/../b              -> b          (still worked)
+    /a/../b             -> /a/../b    (BROKEN: should be /b)
+    a/b/../c            -> a/b/../c   (BROKEN: should be a/c)
+    x/y/../../z         -> x/y/../../z(BROKEN: should be z)
+
+We exercise it through a merge-filter filename, one of the callers that pass
+CFN_COLLAPSE_DOT_DOT_DIRS.  The filter file lives at "a/c"; we reference it as
+"a/none/../c" where the "none" directory does NOT exist.  A correct collapse
+turns that into "a/c" (which opens), so the filter is read and applied; the
+buggy code leaves the literal "a/none/../c", which fails to open (ENOENT) and
+aborts the whole run.  This is also a faithful reproduction of the user-facing
+regression: merge/exclude files referenced via a path containing "..".
+"""
+
+import os
+
+from rsyncfns import (
+    FROMDIR, TODIR, rmtree, run_rsync, test_fail,
+)
+
+rmtree(FROMDIR)
+rmtree(TODIR)
+FROMDIR.mkdir(parents=True, exist_ok=True)
+TODIR.mkdir(parents=True, exist_ok=True)
+
+# Source: one file to keep, one the filter should drop.
+(FROMDIR / 'keep.txt').write_text('keep\n')
+(FROMDIR / 'drop.skip').write_text('drop\n')
+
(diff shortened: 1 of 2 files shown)


dr-who:
I ran mutation testing over this area to check the fix is actually pinned down, and it found a genuine weakness — **in the test, not the code**. Pushed `e42f5917` to close it.

**Method.** Fifteen small edits to `clean_fname()`'s `".."` logic (flip a comparison, shift a bound, move an assignment by one), each rebuilt and run against the suite. Ten survived. To tell a real gap from an *equivalent* mutant I classified each against a reference model of the intended semantics, so "survived" only counts when the output genuinely changes.

**What survived that shouldn't have.** Six of the ten changed behaviour and nothing caught them, including two that matter a lot here:

- restoring the pre-fix `limit = name - 1` initialiser — i.e. **a partial revert of this very fix** — still passed;
- turning `s == name` into `s != name`, which both stops the collapse *and* reintroduces the out-of-bounds read that 9e089846 set out to remove.

**Why.** `a/none/../c` only exercises a `".."` that eats a component with a `/` in front of it. It never reaches the `s == name` branch (collapsing the *leading* component), which is exactly where those mutants live.

**Fix.** Two more merge-filter names in the same test, covering the other two shapes:

- `zzz/../f1` — eats the leading component, where the back-up reaches the start of the name with no earlier `/` to find;
- `../../updir` (run from two directories down) — a leading `".."` has nothing to collapse into and must be *preserved*.

Each name resolves only if that shape is handled exactly right; otherwise the merge file can't be opened, or a different file is opened and `*.skip` isn't filtered out.

Kill rate on behaviour-changing mutants: **5/11 → 10/11**. Of the five still surviving, **four are equivalent mutants** (the model confirms they alter no output), so no test can kill them. The last one — advancing the scan by one byte instead of two — only shows on a name ending in `..`, which can't be a merge-filter file, so I left it uncovered rather than contorting the test to reach it.

Separately, on the **OpenBSD red**: that job timed out after 300s in `daemon-munge`, which I don't think this change can reach — `daemon-munge_test.py` contains no `..`, `filter`, `merge` or `exclude`, the daemon path uses `sanitize_path()` rather than the `..`-collapsing `clean_fname()`, and the test passes 3/3 locally over the same TCP transport. OpenBSD is green on my other three PRs. This push re-triggers it, so we'll see whether it reproduces.

PR #1034, opened 2026-07-21
syscall: let secure_relative_open() fallback create a missing final component

### LINKED ISSUE #1033: secure_relative_open() portable fallback cannot create a new final component (ENOENT vs ENOTDIR). Daemon --inplace fails to create files on kernels without openat2.
### Affected code

`syscall.c`, `secure_relative_open()` — the **per-component `O_NOFOLLOW` walk fallback** (the tier used when there is no kernel `RESOLVE_BENEATH`: NetBSD/OpenBSD/Solaris/Cygwin, **Linux < 5.6** where `openat2` returns `ENOSYS` at runtime, and `--disable-openat2` builds).

Affected versions: current `master` (reproduced on `3.4.3-155-g3b84610c`), v3.4.4, v3.4.3, v3.4.1-sec-patches1. (3.4.0–3.4.2 carry the same walker but no `O_CREAT` caller, so they don't hit it in practice.) Distro backports of the CVE-2026-29518 hardening onto older rsync are also affected (e.g. Red Hat `rsync-3.1.3-26.el8_10`/`-27.el8_10`).

### What happens

The walk probes each component with `openat(dirfd, part, O_RDONLY | O_DIRECTORY | O_NOFOLLOW)` and only falls back to opening the component as a file when the probe fails with **`ENOTDIR`** (final component already exists as a non-directory):

```c
for (const char *part = strtok(path_copy, "/"); part != NULL; part = strtok(NULL, "/")) {
	int next_fd = openat(dirfd, part, O_RDONLY | O_DIRECTORY | O_NOFOLLOW);
	if (next_fd == -1 && errno == ENOTDIR) {
		if (strtok(NULL, "/") != NULL) { errno = ELOOP; goto cleanup; }
		retfd = openat(dirfd, part, flags | O_NOFOLLOW, mode);   /* file fallback */
		goto cleanup;
	}
	if (next_fd == -1) {
		goto cleanup;          /* <-- ENOENT (a to-be-created final file) lands HERE */
	}
	...
}
```

When the final component **does not exist yet** and the caller passed `O_CREAT`, the `O_DIRECTORY` probe fails with **`ENOENT`** (not `ENOTDIR`), so the file fallback is never reached and the function returns `-1`/`ENOENT`. **On the portable fallback, `secure_relative_open()` can never create a new file** — while the `openat2(RESOLVE_BENEATH)` / `O_RESOLVE_BENEATH` fast paths pass `O_CREAT` straight to the kernel and create it fine, so the tiers disagree.

### Real world impact

Since the CVE-2026-29518 hardening (f1c24ab0), the non-chroot daemon receiver routes its `--inplace` destination open through this helper (`receiver.c`):

```c
if (use_secure_symlinks)        /* am_daemon && !am_chrooted */
	fd2 = secure_basis_open(NULL, fnametmp, O_WRONLY|O_CREAT, 0600);
```

So on any portable-fallback platform, a daemon module with `use chroot = no` receiving `--inplace` **fails for every file that does not already exist**.

Real-world casualty: MariaDB/Galera rsync SST on RHEL 8 (kernel 4.18, no `openat2`, distro backport of the same hardening). The joiner's datadir is empty, so every table file is a create; rsync exits 23, the SST script exits 255, Pacemaker leaves the galera resource `FAILED`, and the node can never join. Confirmed by `strace` on the affected host:

```
openat2(AT_FDCWD, "<table>.frm", {O_WRONLY|O_CREAT, RESOLVE_BENEATH}) = -1 ENOSYS
openat (AT_FDCWD, "<table>.frm", O_RDONLY|O_NOFOLLOW|O_DIRECTORY)      = -1 ENOENT
```

### Reproduction (any kernel, current master)

```sh
./configure --disable-openat2     # forces the portable resolver tier;
make                              # Linux <5.6 takes the same path via runtime ENOSYS

cat > /tmp/r.conf <<EOF
port = 8730
address = 127.0.0.1
[sst]
  path = /tmp/dst/
  use chroot = no
  read only = no
EOF
mkdir -p /tmp/dst/db /tmp/src/db
echo hi > /tmp/src/db/new.frm            # a file that does NOT exist on dst yet
./rsync --daemon --no-detach --config /tmp/r.conf &

./rsync --inplace --recursive /tmp/src/db/ rsync://127.0.0.1:8730/sst/db
```

Result:

```
rsync: [receiver] open "/db/new.frm" (in sst) failed: No such file or directory (2)
rsync error: some files/attrs were not transferred (see previous errors) (code 23)
```

Dropping `--inplace` succeeds (temp-file + rename path); re-running `--inplace` once the file exists succeeds — the failure is specifically *create* through the walk.

### Proposed fix

When the `O_DIRECTORY` probe fails with `ENOENT` on the **last** component and the caller wants a file (not `O_DIRECTORY`), open it directly with `flags | O_NOFOLLOW` — mirroring the existing `ENOTDIR` last-component fallback, and making the portable tier consistent with the kernel fast paths (which already honor `O_CREAT`). A missing **intermediate** component (more path follows) stays a genuine `ENOENT`.

```diff
 		if (next_fd == -1) {
+			/* Final component that does not exist yet: if the caller
+			 * wants a file (not O_DIRECTORY), open/create it here with
+			 * O_NOFOLLOW so O_CREAT works and a pre-planted symlink at
+			 * the name is still refused.  A missing *intermediate*
+			 * component (more path follows) stays a genuine ENOENT. */
+			if (errno == ENOENT && !(flags & O_DIRECTORY)
+			 && strtok(NULL, "/") == NULL) {
+				retfd = openat(dirfd, part, flags | O_NOFOLLOW, mode);
+			}
 			goto cleanup;
 		}
```

Security properties are unchanged: intermediate components are still opened `O_DIRECTORY|O_NOFOLLOW` (no symlink traversal, stays beneath basedir); the front-door `..`/absolute-path rejection is untouched; the created final file uses `O_NOFOLLOW`, so a symlink raced into that name yields `ELOOP` rather than an escape (`open(O_CREAT|O_NOFOLLOW)` on an existing symlink does not follow it). Callers without `O_CREAT` opening a missing file still get `ENOENT`.

Validation performed:

- With the patch, the reproduction above succeeds (create + in-place update), and a pre-planted symlink at the destination name is still refused — the symlink target is untouched.
- Full `make check` on the patched `--disable-openat2` build: overall result 0.
- Standalone reproducers (shipped-vs-fixed walker in C, and a syscall-faithful Python model) confirm: buggy walker → `ENOENT` on create; fixed → create OK, existing-file open OK, symlink final → `ELOOP`, `../` escape → `EINVAL`.

Reproducers and the patch are attached (zip). 
[repro_and_patch.zip](https://github.com/user-attachments/files/30238614/repro_and_patch.zip)

### Disclaimer

AI was used in the creation and verification of this bug report

mabunemeh:
Fixes #1033

The per-component O_NOFOLLOW walk fallback in secure_relative_open() -- the tier used when no kernel RESOLVE_BENEATH is available (NetBSD, OpenBSD, Solaris, Cygwin, Linux < 5.6 where openat2 returns ENOSYS, and --disable-openat2 builds) -- probes each component with

	openat(dirfd, part, O_RDONLY | O_DIRECTORY | O_NOFOLLOW);

and only falls back to opening the component as a file when the probe fails with ENOTDIR, i.e. only when the final component already exists as a non-directory.  A final component that does not exist yet fails the probe with ENOENT, which is not special-cased, so the walk returns -1/ENOENT: on this tier secure_relative_open() can never create a new file, no matter what flags the caller passed.

Impact: since the CVE-2026-29518 hardening the
non-chroot daemon receiver routes its --inplace destination open through this helper with O_WRONLY|O_CREAT (receiver.c, secure_basis_open), so every --inplace transfer of a new file into a "use chroot = no" module fails with

	rsync: [receiver] open "..." failed: No such file or directory (2)

and exit code 23 on the fallback tier.  The common real-world casualty is MariaDB/Galera rsync SST on RHEL 8 (kernel 4.18, no openat2, distro backport of the same hardening): the joiner datadir is empty, every table file is a create, and the node can never join.  Reproducible on any kernel with a --disable-openat2 build:

	rsync --daemon (use chroot = no) + rsync --inplace -r src/ dst
	-> fails for every file that does not already exist.

Fix: when the O_DIRECTORY probe fails with ENOENT on the LAST component and the caller wants a file (not O_DIRECTORY), open it directly with the caller flags | O_NOFOLLOW, mirroring the existing ENOTDIR last-component fallback.  O_CREAT now works; a symlink raced into the name is still refused with ELOOP (O_NOFOLLOW); a missing INTERMEDIATE component (more path follows) still fails with ENOENT; the all-components-were-directories and O_DIRECTORY handling is unchanged, as are the openat2/O_RESOLVE_BENEATH fast paths.

### DIFF
--- a/syscall.c
+++ b/syscall.c
@@ -1995,6 +1995,15 @@ int secure_relative_open(const char *basedir, const char *relpath, int flags, mo
 			goto cleanup;
 		}
 		if (next_fd == -1) {
+			/* Final component that does not exist yet: if the caller
+			 * wants a file (not O_DIRECTORY), open/create it here with
+			 * O_NOFOLLOW so O_CREAT works and a pre-planted symlink at
+			 * the name is still refused.  A missing *intermediate*
+			 * component (more path follows) stays a genuine ENOENT. */
+			if (errno == ENOENT && !(flags & O_DIRECTORY)
+			 && strtok(NULL, "/") == NULL) {
+				retfd = openat(dirfd, part, flags | O_NOFOLLOW, mode);
+			}
 			goto cleanup;
 		}
 		if (dirfd != AT_FDCWD) close(dirfd);
--- a/t_secure_relpath.c
+++ b/t_secure_relpath.c
@@ -21,6 +21,18 @@
  * post-fix, the front-door check catches every variant up front
  * with a consistent EINVAL across platforms.
  *
+ * Run with a second argument of "semantics" it instead checks the
+ * resolution semantics that both resolver tiers (the kernel
+ * RESOLVE_BENEATH fast paths and the per-component O_NOFOLLOW walk
+ * fallback) must agree on: creating a missing final component with
+ * O_CREAT, ENOENT for a missing intermediate component, opening an
+ * existing file, refusing an out-of-tree symlink in the final
+ * component, and O_DIRECTORY requests. The create case is a
+ * regression test for the fallback returning ENOENT instead of
+ * creating the file (which broke the non-chroot daemon receiver
+ * with --inplace); the fallback tier is exercised natively on
+ * NetBSD/OpenBSD/Solaris/Cygwin and by --disable-openat2 builds.
+ *
  * Not linked into rsync itself.
  */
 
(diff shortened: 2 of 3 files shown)


steadytao (changes requested):
The source change is narrowly plausible but this resolver is security-sensitive so it needs an automated regression test before merge.

Please cover a missing final component with `O_CREAT`, a missing intermediate component, an existing regular file, a final symlink and an `O_DIRECTORY` request.

mabunemeh (commented):
Added in 41c3aea: a semantics mode for the existing t_secure_relpath helper plus a testsuite/secure-open-semantics_test.py driver covering the five requested cases — missing final component with O_CREAT (single- and multi-component, created file asserted present), missing intermediate component (ENOENT, nothing created), existing regular file (O_RDONLY and O_WRONLY|O_CREAT reopen), a final-component symlink, and O_DIRECTORY requests (existing dir opens and is a dir; missing name stays ENOENT).

Validation: the test passes on both an openat2 build and a --disable-openat2 build, and make check on the --disable-openat2 build is fully green. With the fix reverted, exactly the two create checks fail with ENOENT and the rest still pass, so the test isolates this regression. The fallback tier also runs natively on the NetBSD/OpenBSD/Solaris/Cygwin CI jobs.

steadytao (approved):
LGTM.

PR #1036, opened 2026-07-30
rsync-ssl: -verify_ip when connecting to an IP/IPv6 address

UTsweetyfish:
Fixes:
1. ipcert (Let's Encrypt supports this earlier this year, see [their post](https://letsencrypt.org/2026/01/15/6day-and-ip-general-availability))
```
$ rsync-ssl rsync://192.0.2.1
verify depth is 4
Connecting to 192.0.2.1
depth=0
verify error:num=62:hostname mismatch
40F13FDC967F0000:error:0A000086:SSL routines:tls_post_process_server_certificate:certificate verify failed:../ssl/statem/statem_clnt.c:2103:
rsync: did not see server greeting
rsync error: error starting client-server protocol (code 5) at main.c(1850) [Receiver=3.4.0]
```

2. `openssl s_client -connect 2001:db8::1:874` didn't got properly square-bracketed
```
$ rsync-ssl 'rsync://[2001:db8::1]'
verify depth is 4
s_client: -connect argument or target parameter malformed or ambiguous
rsync: did not see server greeting
rsync error: error starting client-server protocol (code 5) at main.c(1850) [Receiver=3.4.0]
```

```
/usr/bin/openssl s_client [...] -connect 2001:db8::1:874
```

My IPv6 connection is broken so I havn't tested the second case. But it should work.

### DIFF
--- a/rsync-ssl
+++ b/rsync-ssl
@@ -137,7 +137,19 @@ function rsync_ssl_helper {
     fi
 
     if [[ $RSYNC_SSL_TYPE == openssl ]]; then
-	exec $RSYNC_SSL_OPENSSL s_client $caopt $certopt $keyopt -quiet -verify_quiet -servername $hostname -verify_hostname $hostname -connect $hostname:$port
+	if echo "$hostname" | grep -qE '^[0-9]+\.[0-9]+\.[0-9]+\.[0-9]+$|^[0-9a-fA-F:]+$'; then
+	    verify_opt="-verify_ip $hostname"
+	    if echo "$hostname" | grep -q ':'; then
+		# IPv6
+		connect="[$hostname]:$port"
+	    else
+		connect="$hostname:$port"
+	    fi
+	else
+	    verify_opt="-verify_hostname $hostname"
+	    connect="$hostname:$port"
+	fi
+	exec $RSYNC_SSL_OPENSSL s_client $caopt $certopt $keyopt -quiet -verify_quiet -servername $hostname $verify_opt -connect "$connect"
     elif [[ $RSYNC_SSL_TYPE == gnutls ]]; then
 	exec $RSYNC_SSL_GNUTLS --logfile=/dev/null $gnutls_cert_opt $gnutls_key_opt $gnutls_opts $hostname:$port
     else


rsync-ssl
```diff
@@ -137,7 +137,19 @@ function rsync_ssl_helper {
     fi
 
     if [[ $RSYNC_SSL_TYPE == openssl ]]; then
-	exec $RSYNC_SSL_OPENSSL s_client $caopt $certopt $keyopt -quiet -verify_quiet -servername $hostname -verify_hostname $hostname -connect $hostname:$port
+	if echo "$hostname" | grep -qE '^[0-9]+\.[0-9]+\.[0-9]+\.[0-9]+$|^[0-9a-fA-F:]+$'; then
```
steadytao, line 140: This pattern accepts invalid address-like values such as 999.999.999.999 and does not cover all valid IPv6 forms. Would be better to use strict IP parsing rather than classifying addresses with regex.

rsync-ssl
```diff
@@ -137,7 +137,19 @@ function rsync_ssl_helper {
...
+		# IPv6
+		connect="[$hostname]:$port"
+	    else
+		connect="$hostname:$port"
+	    fi
+	else
+	    verify_opt="-verify_hostname $hostname"
+	    connect="$hostname:$port"
+	fi
+	exec $RSYNC_SSL_OPENSSL s_client $caopt $certopt $keyopt -quiet -verify_quiet -servername $hostname $verify_opt -connect "$connect"
```
steadytao, line 152: This still omits `-verify_return_error` so `s_client` may continue after a certificate verification error. IP literals should also use `-verify_ip` without sending an IP address through `-servername`

Please add hostname, IPv4, IPv6, malformed-address and certificate-mismatch coverage.

steadytao (changes requested):
The -verify_ip direction is right but this does not yet make rsync-ssl fail closed or handle IP literals robustly.

rsync-ssl
UTsweetyfish: rsync-ssl is just a simple wrapper for rsync, it will fail in rsync anyway.

btw handling addresses could be tricky, this regex seems can't handle some IPv6 addresses like link-local ones: `fe80::1234%eth0`. idk how to handle this, or whether we should handle this. Any advice?

rsync-ssl
UTsweetyfish: `-verify_return_error` is already in `$caopt`.

rsync-ssl
UTsweetyfish:
```sh
$ rsync -V
rsync  version 3.4.0  protocol version 32
...
$ rsync -vvv -e echo 'rsync://[fe80::1%eno1]:874/module'
opening connection using: echo "fe80::1%eno1" rsync --server --daemon .  (6 args)
...
```

rsync-ssl
steadytao: oh yeah it is, thank you 🤣

rsync-ssl
steadytao: hmmmmmm

steadytao: I feel that since rsync accepts scoped IPv6, I think the wrapper should handle it. There is no need to implement a complete IPv6 regex though. A colon unambiguously selects the IPv6 path then the zone can be removed only from the certificate identity while retained for the connection:

```bash
verify_ip=${hostname%%%*}
connect="[$hostname]:$port"
```

Use `-verify_ip "$verify_ip"` without `-servername`; an IP SAN cannot contain the interface zone. OpenSSL can reject malformed IPv6 itself. IPv4 is the only case that needs a small strict four-octet/range check rather than the current shape-only regex.

PR #1038, opened 2026-08-01
Fix two minor logical bugs

sshedi:
- match.c: Guard against an out-of-bounds array access in match_sums()
- generator.c: Remove dead code in atomic_create()


steadytao (approved):
LGTM. @tridge

PR #1047, opened 2026-08-13
rsync-ssl, testsuite: accept --type=SSL_TYPE anywhere in the args

Zeno-sole:
The --type=... option was only recognized as the first argument, so a command such as "rsync-ssl --dry-run --type=stunnel host::mod" passed the option through to the underlying rsync, which rejected it as unknown.  Scan the full argument list for --type=..., export RSYNC_SSL_TYPE, and drop the option before handing the remaining args to rsync.  The manpage no longer says the option must be first.

Adds a test that runs rsync-ssl with a fake rsync in PATH and checks that --type= is consumed in first, middle, and last positions.

### DIFF
--- a/rsync-ssl
+++ b/rsync-ssl
@@ -226,9 +226,14 @@ if [[ "$1" == --HELPER ]]; then
     rsync_ssl_helper "${@}"
 fi
 
-if [[ "$1" == --type=* ]]; then
-    export RSYNC_SSL_TYPE="${1/--type=/}"
-    shift
-fi
+args=()
+for arg in "$@"; do
+    if [[ "$arg" == --type=* ]]; then
+        export RSYNC_SSL_TYPE="${arg#--type=}"
+    else
+        args+=("$arg")
+    fi
+done
+set -- "${args[@]}"
 
 rsync_ssl_run "${@}"
--- a/rsync-ssl
+++ b/rsync-ssl
@@ -227,8 +227,14 @@ if [[ "$1" == --HELPER ]]; then
 fi
 
 args=()
+dash_dash_seen=false
 for arg in "$@"; do
-    if [[ "$arg" == --type=* ]]; then
+    if [[ $dash_dash_seen == true ]]; then
+        args+=("$arg")
+    elif [[ "$arg" == "--" ]]; then
+        args+=("$arg")
+        dash_dash_seen=true
+    elif [[ "$arg" == --type=* ]]; then
         export RSYNC_SSL_TYPE="${arg#--type=}"
     else
         args+=("$arg")
--- a/testsuite/rsync-ssl-type-option_test.py
+++ b/testsuite/rsync-ssl-type-option_test.py
@@ -8,9 +8,11 @@
 # passed through to rsync verbatim, so an operand such as `--type=stunnel`
 # that was protected from option parsing is not consumed by the wrapper.
 #
-# A fake rsync in PATH records the args it receives; rsync-ssl is run in its
-# normal (non-HELPER) mode with --type= in various positions and the recorded
-# args must contain every other option but never a --type= token.
+# A fake rsync in PATH records the args it receives and the RSYNC_SSL_TYPE it
+# observes; rsync-ssl is run in its normal (non-HELPER) mode with --type= in
+# various positions.  The recorded args must contain every other option but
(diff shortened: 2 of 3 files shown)


Zeno-sole:
**Before repair：**
RSYNC_SSL_CA_CERT=/tmp/server.crt rsync-ssl --dry-run --type=stunnel 10.20.8.153::backuprsync: --type=stunnel: unknown option
rsync error: syntax or usage error (code 1) at main.c(1806) [client=3.4.4]

**After repair：**
RSYNC_SSL_CA_CERT=/tmp/server.crt rsync-ssl --dry-run --type=stunnel 10.20.8.153::backup
drwxrwxr-x          4,096 2026/08/13 17:54:16 .
-rw-rw-r--              0 2026/08/13 17:54:16 111

Zeno-sole:
I would like to kindly request a code review from the upstream author/maintainer. @tridge

steadytao:
No need to ping after less then 24 hours. 1) a release has just been made with **major** security concerns, we are resting. 2) its been less then 24 hours :D

Let's be respectful. "kindly" doesn't make something respectful.

rsync-ssl
```diff
@@ -226,9 +226,14 @@ if [[ "$1" == --HELPER ]]; then
...
-fi
+args=()
+for arg in "$@"; do
+    if [[ "$arg" == --type=* ]]; then
+        export RSYNC_SSL_TYPE="${arg#--type=}"
+    else
+        args+=("$arg")
+    fi
+done
+set -- "${args[@]}"
```
steadytao, line 243: Stop interpreting wrapper options at --

As written, `rsync-ssl -- --type=stunnel` consumes an operand that was explicitly protected from option parsing. Preserve -- and every subsequent argument, and add a regression for this case.

steadytao (changes requested):

rsync-ssl
Zeno-sole: Thanks for the feedback — you're right, and I've fixed it.

  The wrapper-option scan now stops at a -- argument: rsync-ssl preserves the -- and every subsequent argument verbatim, passing them
  through to the underlying rsync unchanged. So a --type=... operand that was explicitly protected from option parsing (e.g. rsync-ssl --
  --type=stunnel host::mod) is no longer consumed by the wrapper, while a --type=... before the -- is still recognized as before.

  I also added regression coverage for both cases in testsuite/rsync-ssl-type-option_test.py (a ---protected operand is preserved, and a
  wrapper option before -- is still consumed), and documented the behavior in rsync-ssl.1.md.

testsuite/rsync-ssl-type-option_test.py
```diff
@@ -0,0 +1,80 @@
...
+
+base = SCRATCHDIR / 'rsync-ssl-type-opt'
+rmtree(base)
+base.mkdir(parents=True)
+
+args_capture = base / 'rsync_args'
+fakebin = base / 'bin'
+fakebin.mkdir(parents=True)
+fake_rsync = fakebin / 'rsync'
+fake_rsync.write_text(f"#!/usr/bin/env bash\nprintf '%s\\n' \"$@\" > {args_capture}\nexit 0\n")
```
steadytao: Please capture and assert `RSYNC_SSL_TYPE` as well as argv. The current test proves that `--type=` is removed but it would also pass if the wrapper silently discarded the requested SSL implementation instead of exporting it. `subprocess.run()` should also check the wrappers exit status.

steadytao (changes requested):

testsuite/rsync-ssl-type-option_test.py
Zeno-sole: Thanks for the review. This has been addressed in commit `916b75be4d73d8fb9621647353a33a6d9fd3f907`:

1. The fake rsync now captures both the received arguments and `RSYNC_SSL_TYPE`.
2. Each test case asserts the expected `RSYNC_SSL_TYPE`; cases where `--type=` must not be consumed assert that the variable is unset.
3. The test now checks the `rsync-ssl` exit status and reports the exit code and output on failure.

I also verified that the test fails if the wrapper silently discards `--type=`, or if the fake rsync exits with a non-zero status.

steadytao (approved):
LGTM.

PR #1048, opened 2026-08-13
rrsync: support fd pins in user namespaces

### LINKED ISSUE #1044: [3.5.0 regression] rrsync fails inside a user namespace: "/proc/self/fd/N" ELOOP
I can reproduce the following issue on the Debian package but I asked an AI for a reproducer that doesn't require the Debian package.

I believe the reason I did not spot this earlier is because the embargo patchset didn't carry the runtests.py refactor, but have not double checked.

## What happens

In 3.5.0, transfers through `rrsync` fail when rsync runs inside a user namespace
that does not map global uid 0 — rootless podman/docker, Docker `userns-remap`,
`sbuild --chroot-mode=unshare`, etc:

```
rsync: [Receiver] change_dir#1 "/proc/self/fd/4/" failed: Too many levels of symbolic links (40)
rsync error: errors selecting input/output files, dirs (code 3) at main.c(768) [Receiver=3.5.0]
```

Outside such a namespace the same commands work. This also makes the Debian package
FTBFS, since sbuild's default build mode is one of these namespaces.

Version: 3.5.0, also reproduced on git master (`dcef974c`). x86_64, Linux 7.1.7,
glibc 2.43.

## Reproducing with the testsuite

In a built tree — `make`, plus `make check-progs` for the test helpers — the only
change is the `unshare` prefix:

```
$ unshare --user --map-root-user --mount --pid --fork --mount-proc \
      ./runtests.py --rsync-bin=$PWD/rsync --always-log \
      rrsync-pull-delivers-content rrsync-backup-dir-inband-pivot

----- rrsync-pull-delivers-content log follows
push through rrsync failed (rc=3, output='rsync: [Receiver] change_dir#1 "/proc/self/fd/4/" failed: Too many levels of symbolic links (40)\nrsync error: errors selecting input/output files, dirs (code 3) at main.c(768) [Receiver=3.5.0]')
FAIL    rrsync-pull-delivers-content

----- rrsync-backup-dir-inband-pivot log follows
AssertionError: the transfer failed, but not through the pinned backup dir: an unrelated failure would satisfy the check above: stderr='rsync error: error in file IO (code 11) at rsync.c(900) [receiver=3.5.0]\n'
FAIL    rrsync-backup-dir-inband-pivot

      0 passed
      2 failed
```

Without the prefix both tests pass.

## Reproducing without rrsync

Any `/proc/self/fd/...` path handed to the resolver hits it, so it can be shown with
one rsync invocation:

```sh
cat > /tmp/min.sh <<'EOF'
#!/bin/sh
rm -rf /tmp/r; mkdir -p /tmp/r/dst /tmp/r/src; echo hello > /tmp/r/src/f
echo "  /proc/self is owned by: $(stat -c '%U (uid %u)' /proc/self)"
exec 9< /tmp/r/dst
"$1" -a --confine-root=/tmp/r/dst --log-file=/proc/self/fd/9/rsync.log \
     /tmp/r/src/ /tmp/r/dst/ 2>&1 | sed 's/^/  /'
EOF
chmod +x /tmp/min.sh

echo '### outside a user namespace ###'; /tmp/min.sh ./rsync
echo '### inside a user namespace ###'
unshare --user --map-root-user --mount --pid --fork --mount-proc /tmp/min.sh ./rsync
```

```
### outside a user namespace ###
  /proc/self is owned by: root (uid 0)
### inside a user namespace ###
  /proc/self is owned by: nobody (uid 65534)
  rsync: [client] failed to open log-file /proc/self/fd/9/rsync.log: Too many levels of symbolic links (40)
  Ignoring "log file" setting.
```

The difference between the two runs is the reported owner of `/proc/self`: inside the
namespace global uid 0 has no mapping, so the link reads as the overflow uid, and the
symlink-ownership check in `ona_open()` (`syscall.c`) refuses to traverse it with
ELOOP. Every pin is spelled `/proc/self/...`, so no pin resolves.

## Bisect

`git bisect run` over `v3.4.4..v3.5.0`, testing a push through rrsync under the
`unshare` prefix above:

> **88cee089637ac30f3af7b4797e978fffc2cc45f5** — *rrsync: pin path components against
> a TOCTOU and fail closed on anomalies*

That is where rsync starts being handed `/proc/self/fd/N` paths. The ownership check
they run into came in seven commits earlier, in
`54965efca919996786b19e9bda527e04ee89c334`.

steadytao:
Fixes #1044

## Summary
- allow the exact `/proc/self` component of a recognised rrsync fd pin when procfs reports its owner as the overflow UID inside a user namespace
- retain the normal symlink-owner and confinement checks for every other component and for the resolved fd target
- add a regression test that enters an unprivileged user namespace and exercises a confined `/proc/self/fd/N` log path

### DIFF
--- a/syscall.c
+++ b/syscall.c
@@ -401,9 +401,13 @@ static int ona_open(const char *path, int flags, mode_t mode, char *out_abs, siz
 		}
 
 		if (S_ISLNK(lst.st_mode)) {
-			/* Symlink: untrusted owner is refused; trusted owner
-			 * is followed via readlinkat + splice. */
-			if (lst.st_uid != 0 && lst.st_uid != trusted_uid) {
+			/* Symlink: untrusted owner is refused; trusted owner is followed
+			 * via readlinkat + splice.  In a user namespace procfs reports
+			 * /proc/self with the overflow uid, so allow that exact component
+			 * while traversing a recognised fd pin. */
+			int proc_self_pin = pin_transit && strcmp(abspath, "/proc") == 0
+					  && strcmp(comp, "self") == 0;
+			if (!proc_self_pin && lst.st_uid != 0 && lst.st_uid != trusted_uid) {
 				saved_errno = ELOOP;
 				goto out;
 			}
--- a/syscall.c
+++ b/syscall.c
@@ -143,12 +143,17 @@ static const char *confinement_root(unsigned int *lenp)
 	return confine_root;
 }
 
-/* Split the "/proc/<self|pid>/fd" prefix off `p`, returning the tail -- "" for
- * the pin directory itself, otherwise a string starting with '/'.  NULL when `p`
- * is not in the fd-pin namespace at all. */
+/* Split a recognised fd-pin prefix off `p`, returning the tail -- "" for the
+ * pin directory itself, otherwise a string starting with '/'.  NULL when `p`
+ * is not in an fd-pin namespace. */
 static const char *fd_pin_tail(const char *p)
 {
 	const char *s;
+
+	if (strncmp(p, "/dev/fd", 7) == 0) {
+		s = p + 7;
+		return (*s == '\0' || *s == '/') ? s : NULL;
+	}
 
 	if (strncmp(p, "/proc/", 6) != 0)
 		return NULL;
@@ -168,8 +173,8 @@ static const char *fd_pin_tail(const char *p)
 	return (*s == '\0' || *s == '/') ? s : NULL;
 }
 
-/* An EXACT pin entry, "/proc/self/fd/7" -- the one spelling whose target is what
- * confinement must judge.  rrsync also writes a pinned parent as
+/* An EXACT pin entry, such as "/proc/self/fd/7" or "/dev/fd/7", whose target is
+ * what confinement must judge.  rrsync also writes a pinned parent as
  * ".../fd/7/<leaf>", but the walk resolves the magic link itself and checks the
  * components past it, so only the bare entry is resolved here.  Requiring all
  * digits keeps a planted name like ".../fd/outside-secret" out. */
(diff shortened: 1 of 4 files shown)


tridge:
@samueloph please test

samueloph:
@tridge @steadytao I confirm this solves the problem on Debian

seks99x:
@steadytao Great job on this fix! I did notice one small edge case during testing, if we change the test to use `/dev/fd/` instead of `/proc/self/fd/`, the test still fails. It looks like `/dev/fd/` will require some extra handling or allowlisting alongside `/proc/self/fd/`


```
----- new log follows
confined log path through an fd pin was rejected (rc=0, stderr='rsync: [client] failed to open log-file /dev/fd/3/rsync.log: Too many levels of symbolic links (40)')
user-namespace regression test failed (rc=1)
----- new log ends
FAIL    fd_test
------------------------------------------------------------
----- overall results:
      0 passed
      1 failed
------------------------------------------------------------
overall result is 1
```

steadytao:
Thanks... /dev/fd reaches a different kernel magic-link path. I would keep this change limited to the user-namespace /proc/self ownership exception rather than broaden the confined path traversal here.

steadytao:
Just realised that #1054 would probably cover it anyways

steadytao:
Yes it would, should be good to merge this.

seks99x:
Actually I don't think this is covered in #1054 because I didn't modify any permission changes but we could fix that on the same PR if you would like.

steadytao:
Ah, yeah, I just assumed from 1053's scope. Nonetheless, I don't fully see why we need to broaden the confinement exception to /dev/fd/N seeing as those pins are never generated?

seks99x:
Im not with broadening as what you saying. The issue is Bash process substitution uses /dev/fd. If the same case happened using Bash substitution it will mostly fail with ELOOP. Some other cases we may be missing (which might be ok) but I think if we will already make exceptions to broaden the checks more we would add /dev/fd & /proc/self/ only as they are the most common. What do you think?

steadytao:
Ah yeah true, didn't think of those cases. Will broaden a bit. Thank you, Omar.

PR #1062, opened 2026-08-19
testsuite: probe the punch granularity each hole assertion relies on

samueloph:
The last section of preallocate_test.py writes 256 blocks of [4 KiB data][24 KiB zeros][4 KiB data] and requires --inplace --sparse to deallocate at least half of the file.  A punch only frees whole allocation units, and that interior run sits 4 KiB into each 32 KiB block: with a 4 KiB unit it covers six whole units (24 KiB freed per block), but with a 16 KiB unit it covers none at all -- [0,16K) holds live data at [0,4K) and [16K,32K) holds live data at [28K,32K) -- so st_blocks cannot move whatever rsync does.

That fails the build on Debian loong64, whose kernel uses 16 KiB pages:

    --inplace --sparse left matching interior zero runs allocated: 8388608 of
    8388608 bytes remain allocated after a 8388608-byte matched-block update

i.e. exactly "nothing was freed".  fs_can_punch_holes() does not catch it because it punches a whole 64 KiB file, which frees blocks at any granularity.

Generalise that helper into punch_frees(offset, length, size) so each assertion can probe the shape it actually depends on, and gate the interior one on the exact layout written just above it.  Where the filesystem can free that run the assertion runs exactly as before; where it cannot, the test says so rather than reporting a regression.  The content check is unconditional either way.

While here, call fallocate64() instead of fallocate().  ctypes declares the offset and length as c_longlong, but glibc's fallocate() takes 32-bit off_t where off_t is 32 bits, so the callee reads the high half of `offset` as its `length`, the call fails with EINVAL, and can_punch comes back False -- which is why none of this file's hole-punching assertions have ever run on a 32-bit port.  On i386:

    fs_can_punch_holes() as written : before=128 ret=-1 errno=22 -> False
    the same probe via fallocate64  : before=128 ret=0            -> True

fallocate64() takes off64_t on every architecture and is a plain alias of fallocate() where off_t is already 64 bits.  The probe also writes urandom rather than a repeated byte, so a filesystem that compresses does not answer "nothing was freed" merely because nothing was allocated.

Failing build log:
  loong64 https://buildd.debian.org/status/fetch.php?pkg=rsync&arch=loong64&ver=3.5.0%2Bds1-1&stamp=1786938228&raw=0

### DIFF
--- a/testsuite/preallocate_test.py
+++ b/testsuite/preallocate_test.py
@@ -38,12 +38,33 @@ if run_rsync('-a', '--preallocate', f'{src}/', f'{TODIR}/',
              check=False, capture_output=True).returncode != 0:
     test_skipped("--preallocate not supported on this platform")
 
-def fs_can_punch_holes():
-    """True only where the kernel can deallocate blocks via FALLOC_FL_PUNCH_HOLE
-    -- the mechanism do_punch_hole uses for --sparse. A filesystem may report
-    seek-based sparseness yet still keep every block on a punch (e.g. where
-    rsync's punch falls back to writing zeros), so probe the real capability and
-    assert the hole only where it actually frees blocks."""
+def punch_frees(offset, length, size):
+    """True where punching [offset, offset+length) out of a `size`-byte file
+    really deallocates blocks -- the mechanism do_punch_hole uses for --sparse.
+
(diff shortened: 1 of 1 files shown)


steadytao (approved):
LGTM. Thank you for the PR.

PR #1063, opened 2026-08-19
testsuite: interpose the large-file spellings of open/openat/fstatat

samueloph:
Which symbol names this test's LD_PRELOAD hook exports is decided by the compiler that builds it, and which names rsync imports is decided by configure -- and the two do not have to agree.

Where off_t is not already 64 bits, configure's AC_SYS_LARGEFILE adds -D_FILE_OFFSET_BITS=64 (i386 and alpha in Debian), so glibc redirects every open()/openat()/fstatat() call in rsync to open64(), openat64() and fstatat64().  The hook is compiled by a bare "cc", so on those architectures it defines only the unsuffixed names: the receiver's opens never reach it, no EACCES is injected, the marker the positive control looks for is never written, and the test fails with

    positive control failed: receiver did not open the existing partial file
    with O_CREAT (rc=0, output='')

It happens to work on Debian's 64-bit time_t ports (armhf, hppa, powerpc, ...) only by luck: their gcc predefines -D_FILE_OFFSET_BITS=64 -D_TIME_BITS=64, so glibc's __REDIRECT renames the hook's own DEFINITIONS as well and it ends up exporting exactly the *64 names rsync imports.

Define both spellings explicitly so the hook interposes whichever set the rsync under test was linked against, and #undef the two macros at the top of the hook so that renaming cannot happen -- otherwise, on precisely those ports whose compiler predefines them, open() would be emitted as open64() and collide with the explicit wrapper ("symbol `open64' is already defined"), leaving the hook unbuildable and the test skipped.

The new pointers are resolved through hook_resolve(), so the existing nested-dlsym recursion guard covers them as well.

Failing build logs:
  i386  https://buildd.debian.org/status/fetch.php?pkg=rsync&arch=i386&ver=3.5.0%2Bds1-1&stamp=1786938290&raw=0
  alpha https://buildd.debian.org/status/fetch.php?pkg=rsync&arch=alpha&ver=3.5.0%2Bds1-1&stamp=1786973432&raw=0

### DIFF
--- a/testsuite/partial-protected-regular-retry-linux_test.py
+++ b/testsuite/partial-protected-regular-retry-linux_test.py
@@ -37,6 +37,17 @@ if _proto is not None and _proto < 30:
 
 hook_code = r'''
 #define _GNU_SOURCE
+/* Build the hook itself WITHOUT large-file redirection, whatever the compiler
+ * defaults to.  Debian's armhf/hppa/powerpc gcc predefines
+ * -D_FILE_OFFSET_BITS=64 -D_TIME_BITS=64 (check with "gcc -v -E -"), and under
+ * those macros glibc's __REDIRECT renames the DEFINITIONS below -- open()
+ * becomes open64(), fstatat() becomes __fstatat64_time64() -- which then
+ * collide with the explicit large-file wrappers further down ("symbol `open64'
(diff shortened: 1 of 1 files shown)


steadytao (approved):
LGTM. Thank you.

PR #1065, opened 2026-08-19
syscall: use O_PATH for held directory traversal

### LINKED ISSUE #1064: rsync still requires read permission for search-only directories
On Linux, rsync still fails to copy an explicitly named file below a directory that is searchable but not readable.

```sh
tmp=$(mktemp -d)
mkdir -p "$tmp/src/xonly" "$tmp/dst"
printf 'test\n' > "$tmp/src/xonly/file"
chmod 111 "$tmp/src/xonly"
(cd "$tmp/src" && rsync -aR xonly/file "$tmp/dst/")
```

This fails with:

```text
rsync: [sender] send_files failed to open ".../src/xonly/file": Permission denied (13)
```

I would expect the named file to be copied. The process can search `xonly` and read the file, and this operation does not need to list the directory.

#1052 fixed the same permission issue in `ona_open()`, but the separate held-directory-fd resolver still opens traversal and parent descriptors with `O_RDONLY | O_DIRECTORY`. The same issue also affects known-file creation below writable/searchable but unreadable destination directories.

Reproduced on rsync master at 0145b912.

I have a patch and regression test ready for this path.

FusionPow:
Use `O_PATH` on Linux for directory descriptors used only for path traversal, `fchdir()`, or as `*at()` parents. Reopen final directory endpoints with the caller-requested access mode and retain the existing `O_RDONLY` fallback elsewhere.

Add coverage for exact sources and readable descendants beneath search-only directories, known-file creation beneath write/search-only destinations, and retained refusal to enumerate unreadable directories.

Fixes #1064.

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -146,28 +146,29 @@ following symlinks by design.
 
 ### The mechanism
 
-Resolution of attacker-influenceable paths goes through `secure_relative_open()`
-and the `do_*_at()` wrappers in `syscall.c`, never a raw `open()`/`rename()`/
-`chmod()` on a full path string. The principle is: **trust the operator-named
-transfer root, and confine all resolution beneath it**, rejecting escapes via
-`..` above the anchor, absolute symlinks, or out-of-tree symlinks.
-`secure_relative_open()` resolves the parent directory by walking it one
-component at a time on a stack of held directory fds, then operates on the final
-component with an at-style call on the resulting directory fd.
+Resolution of attacker-influenceable paths goes through `secure_relative_open()`,
+`secure_relative_dirfd()`, and the `do_*_at()` wrappers in `syscall.c`, never a
+raw `open()`/`rename()`/`chmod()` on a full path string. The principle is:
+**trust the operator-named transfer root, and confine all resolution beneath
+it**, rejecting escapes via `..` above the anchor, absolute symlinks, or
+out-of-tree symlinks. `secure_relative_open()` opens the resolved endpoint with
+the caller's requested access. `secure_relative_dirfd()` instead returns
+traversal authority for `fchdir()` or an at-style operation on a known child;
+it does not imply permission to enumerate the directory.
 
 For per-entry work the receiver and generator go one step further and hold the
 parent directory open: `open_dir_secure()` resolves an entry's directory once
-(via `secure_relative_open()`), `held_dfd_for()` caches that descriptor for the
-duration of the entry, and every operation on the entry — `lstat`, the temp-file
-`mkstemp`, the temp->final `rename`, `chmod`/`chown`/`utimes`, `mkdir`, special-
-file and symlink creation, the delta-basis open, and the recursive delete — runs
-through that one held fd via an `*at()` call (`do_*_atfd()`). Because the
-descriptor is pinned to the directory inode, a parent component flipped to a
-symlink *after* the open cannot redirect any of those operations. The alternate-
-destination lookups are confined the same way (`basis_link_stat()` in
-`generator.c` and `secure_basis_open()` in `receiver.c`), so a peer-chosen
-`--link-dest`/`--compare-dest`/`--copy-dest` basis index cannot reach an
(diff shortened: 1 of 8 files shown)


seks99x:
This looks good Thanks!

I think there is still some cases like this not covered in the source which I still didn't confirm, which if confirmed, we should step back and find a broader solution that covers all edge cases cleanly ( like fixing  `ds_take()` itself or something ) rather than stacking patches on top of one another. I'll come back to this again.

steadytao:
It does LGTM. I'll leave this for a bit while @seks99x looks for some possible untouched cases but overall good to merge in about 2-3 days if unconfirmed.

seks99x:
Sorry for the delay. I'll confirm this tomorrow

tridge:
**Automated review note — AI-generated (Claude), validated against the live diff.** Please sanity-check before acting.
Full report (with the other AIReview PR): https://uav.tridgell.net/RsyncReviews/index.html#pr1065

**Reviewed at head `c512980a46` — verdict: COMMENT.** Nicely constructed. The split into `secure_relative_open()` (readable endpoint) vs the new `secure_relative_dirfd()` (`O_PATH` endpoint) is clean, and I traced every consumer of the now-`O_PATH` held fd (generator.c, sender.c, delete.c, rsync.c, util1.c, the `do_*_atfd` shims, the ACL `/proc/self/fd` path): each uses it only for `*at()`, `fchdir()`, or as pathname authority — all legal on an O_PATH dir fd. The one forbidden op, `fdopendir()`, stays on a readable fd (`secure_opendir()` requests `O_RDONLY|O_DIRECTORY`; delete enumeration goes via `get_dirlist()`, not `del_dirfd`). The fallback is real (`O_RDONLY|O_DIRECTORY`, not `#define O_PATH 0`), the security property is unchanged (pinned fds + per-component `O_NOFOLLOW`), and the endpoint-reopen refactor is fd-clean. No blocking defect. Three points:

- **ISSUE (portability) — `fchdir()` on an `O_PATH` fd regresses on Linux 2.6.39–3.4.** `O_PATH` arrived in 2.6.39, but O_PATH descriptors only became usable with `fchdir()` in **3.5**. The held dirfds are now `O_PATH` (`syscall.c:82`) and the confinement path `fchdir()`s them (`util1.c:1258/1351/1369`), so a binary built against current headers and run on a 2.6.39–3.4 kernel gets `EBADF` — daemon-module / receiver-destination setup fails even for ordinary readable dirs (pre-2.6.39, the O_PATH opens themselves fail). FreeBSD is unaffected (its O_PATH permits `fchdir()`). Those kernels are long EOL so impact is low, but it's a silent regression on a security-relevant path; a cheap `EBADF` fallback (reopen `"."` as `O_RDONLY|O_DIRECTORY` relative to the pinned fd and `fchdir()` that) or a documented minimum kernel would close it. (Surfaced by an independent cold pass; verified against source + kernel history.)
- **NOTE — SECURITY.md over-states the non-Linux fallback:** it says other platforms keep the `O_RDONLY` fallback, but the gate is feature-based (`#if defined O_PATH`), so FreeBSD ≥13 also uses O_PATH fds (and supports them for `fchdir()`). Not a regression there — but the BSD path is untested (the new test self-skips off Linux), so a FreeBSD smoke test is worth it and the doc should be corrected.
- **NOTE — SECURITY.md O_NOFOLLOW wording:** it says anchors use `O_PATH|O_DIRECTORY|O_NOFOLLOW`, but anchors are opened *without* `O_NOFOLLOW` (`open_anchor_dirfd`); only traversal components carry it.

(seks99x/steadytao have already LGTM'd; none of the above blocks that.)

seks99x:
Looks great! I'll investigate later to see if there are any other edge cases missing caller flag support, and I'll write the test suite cases for them. None of that needs to hold up the PR, it's good to merge.

steadytao:
Yeah. Happy to merge this one to get the fix out. Elsewhise can be sorted in another PR. Thank you all.

PR #1070, opened 2026-09-01
Add 4 KiB logical block footprint tracking to stats

### LINKED ISSUE #1066: Feature Request: Report unique 4 KiB logical blocks touched by rsync writes (e.g. in --stats)
## Motivation

Rsync's existing transfer statistics report the amount of data transferred, but that alone does not describe the logical write footprint of a delta update.

For workloads such as large database files, VM images, or other files where relatively small changes are scattered across a large file, the number of bytes changed and the number of storage-sized regions touched can be quite different.

For example, a transfer might modify only 1 MiB of data while requiring writes spread across hundreds of separate 4 KiB regions. Reporting only the number of bytes transferred does not make that distinction visible.

## Proposed feature

Add a statistic to `--stats` reporting the number of **unique 4 KiB logical file blocks touched by `rsync`'s own write operations**.

For example:

```
Literal data written:        1.0 MiB
4 KiB logical blocks touched: 256
```

The statistic would count each 4 KiB logical block at most once for a given file, even if `rsync` writes to that block multiple times during the operation.

The purpose is to expose the logical write footprint of the `rsync` operation, complementing the existing byte-based statistics.

## Scope and terminology

This is intentionally a statistic about **rsync's logical writes**. It is not intended to report:

* dirty memory pages
* filesystem block writes
* physical device I/O
* SSD/NAND write amplification
* flash wear

Those quantities depend on filesystem behavior, caching, CoW, compression, delayed allocation, journaling, the block device, and other layers below `rsync`.

The proposed statistic therefore should use terminology such as **"4 KiB logical blocks touched"** or **"4 KiB logical write footprint"**, rather than implying that `rsync` is measuring physical storage writes.

## Implementation considerations

The statistic would be derived from the logical file offsets and lengths of `rsync`'s own write operations.

The implementation should avoid unnecessary memory overhead, particularly for very large files or transfers with a large number of distinct regions. An implementation based on ranges or another compact representation may be preferable to storing every block individually.

The exact implementation is left open to the maintainers.

## Why this is useful

This statistic would make it possible to distinguish transfers that have the same byte count but very different write patterns.

It would also provide a useful storage-oriented complement to the existing `rsync` statistics without requiring `rsync` to make assumptions about what the underlying filesystem or storage device ultimately does with those writes.

seks99x:
Fixes #1066 
Original feature request by @birdie-github

Currently, rsync --stats reports Literal data, which represents the unmatched file-update data sent over the network. However, network payload is often completely disconnected from the actual disk I/O footprint on the receiving end. Depending on the transfer flags (e.g., default temp-file rebuilding vs. --inplace), a tiny network payload can result in massive disk writes, or vice versa. System administrators currently have no native visibility into the true I/O cost of a sync.

This PR introduces a new metric to the --stats output: Number of modified 4K blocks. It tracks exactly how many 4,096-byte boundaries are written to the receiver's disk, completely decoupling network delta payload from disk I/O footprint.
O(1) Zero-Overhead Implementation
To ensure tracking does not bottleneck the receiver's high-speed I/O loop, the block calculation operates entirely in O(1) time with zero system calls and no loop iterations.
Stateful Tracking: It uses a static state machine (last_tracked_fd and last_touched_blk) to remember the last logical block written. Because the receiver strictly writes sequentially, we can calculate block deltas using purely mathematical boundaries.
Math over Memory: For each write, it calculates the block boundaries using (offset % 4096) and len. It diffs this against the last known state and instantly updates the global counter.

Real-World Examples (Why this matters)
Here is a demonstration of how network data and disk I/O diverge, using a 4MB file where we inject a single byte at 10 different 4K block boundaries:

Case 1: Delta transfer with --inplace
```Bash
# 1. Restore the base file to exactly match the scattered destination
cp dest_scattered.bin base.bin

# 2. Inject a single byte at 10 different 4K block boundaries
python3 -c 'f=open("base.bin", "r+b"); [(f.seek(i*4096), f.write(b"\xFF")) for i in range(1,11)]; f.close()'

# 3. Sync it
../rsync -a --stats --inplace -I --no-whole-file base.bin dest_scattered.bin | egrep "Literal data|touched"
Literal data: 20,480 bytes
Number of modified 4K blocks: 10
```
Context: Rsync sends a small amount of literal data over the network to patch the file. Because --inplace is used, the receiver seeks and updates only the 10 specific blocks on disk. Network and Disk I/O are both low.
Case 2: Delta transfer without --inplace (Default temp-file behavior)
```Bash
# 1. Restore the base file to exactly match the scattered destination
cp dest_scattered.bin base.bin

# 2. Inject a single byte at 10 different 4K block boundaries
python3 -c 'f=open("base.bin", "r+b"); [(f.seek(i*4096), f.write(b"\xFF")) for i in range(1,11)]; f.close()'

# 3. Sync it
../rsync -a --stats -I --no-whole-file base.bin dest_scattered.bin | egrep "Literal data|touched" 

Literal data: 0 bytes
Number of modified 4K blocks: 1,024
```
Context: Rsync perfectly matches the unchanged data locally. The network payload is effectively zero. However, because rsync reconstructs the file into a new hidden temp file before moving it into place, the receiver's disk is forced to rewrite the entire 4MB file (1,024 blocks).

### DIFF
--- a/fileio.c
+++ b/fileio.c
@@ -34,6 +34,7 @@
 #define ALIGNED_LENGTH(len) ((((len) - 1) | (ALIGN_BOUNDARY-1)) + 1)
 
 extern int sparse_files;
+extern struct stats stats;
 
 OFF_T preallocated_len = 0;
 
@@ -165,7 +166,7 @@ static int write_sparse(int f, int use_seek, OFF_T offset, const char *buf, int 
 		}
 		if (i > start) {
 			if (!use_seek)
-			   track_block_touches(f, offset + start, i - start);
+				track_block_touches(f, offset + start, i - start);
 			if (emit_sparse_span(f, use_seek, buf + start, i - start) < 0)
 				return -1;
 			sparse_past_write = offset + i;
@@ -210,34 +211,34 @@ int flush_write_file(int f)
 
 void reset_block_tracker(void)
 {
-    last_tracked_fd = -1;
-    last_touched_blk = -1;
+	last_tracked_fd = -1;
+	last_touched_blk = -1;
 }
 
-void track_block_touches(int f, OFF_T offset, int32 len) 
+void track_block_touches(int f, OFF_T offset, int32 len)
 {
-    if (len <= 0) 
-	    return;
-    extern struct stats stats;
-    if (f != last_tracked_fd) {
-        last_tracked_fd = f;
-        last_touched_blk = -1;
-    }
-    int64 start_blk = offset / 4096;
-    int64 end_blk = start_blk + (((offset % 4096) + len - 1) / 4096);
-    int64 blocks_to_add = 0;
-    if (start_blk > last_touched_blk) 
-        blocks_to_add = (end_blk - start_blk + 1);
-    else if (end_blk > last_touched_blk) 
-        blocks_to_add = (end_blk - last_touched_blk); 
-    if (blocks_to_add > 0) {
-        if (INT64_MAX - stats.touched_blocks_4k < blocks_to_add)
-            stats.touched_blocks_4k = INT64_MAX; 
-        else
-            stats.touched_blocks_4k += blocks_to_add;
-    }
-    if (end_blk > last_touched_blk)
-        last_touched_blk = end_blk;
+	int64 start_blk, end_blk, blocks_to_add = 0;
+
+	if (len <= 0)
+		return;
+	if (f != last_tracked_fd) {
+		last_tracked_fd = f;
+		last_touched_blk = -1;
+	}
+	start_blk = offset / 4096;
+	end_blk = start_blk + (((offset % 4096) + len - 1) / 4096);
+	if (start_blk > last_touched_blk)
+		blocks_to_add = end_blk - start_blk + 1;
+	else if (end_blk > last_touched_blk)
+		blocks_to_add = end_blk - last_touched_blk;
+	if (blocks_to_add > 0) {
+		if (INT64_MAX - stats.touched_blocks_4k < blocks_to_add)
+			stats.touched_blocks_4k = INT64_MAX;
+		else
+			stats.touched_blocks_4k += blocks_to_add;
+	}
+	if (end_blk > last_touched_blk)
+		last_touched_blk = end_blk;
 }
 
 /* write_file does not allow incomplete writes.  It loops internally
@@ -247,9 +248,8 @@ int write_file(int f, int use_seek, OFF_T offset, const char *buf, int len)
 {
 	int ret = 0;
 
-    if (!use_seek && sparse_files == 0) {
-        track_block_touches(f, offset, len);
-	}
+	if (!use_seek && sparse_files == 0)
+		track_block_touches(f, offset, len);
 	while (len > 0) {
 		int r1;
 		if (sparse_files > 0) {
--- a/fileio.c
+++ b/fileio.c
@@ -40,6 +40,9 @@ OFF_T preallocated_len = 0;
 static OFF_T sparse_seek = 0;
 static OFF_T sparse_past_write = 0;
 
+static int last_tracked_fd = -1;
+static int64 last_touched_blk = -1;
+
 int sparse_end(int f, OFF_T size, int updating_basis_or_equiv)
 {
 	int ret = 0;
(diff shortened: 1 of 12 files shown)


fileio.c
```diff
@@ -201,13 +205,44 @@ int flush_write_file(int f)
 	return ret;
 }
 
+void track_block_touches(int f, OFF_T offset, int32 len) 
+{
+    if (len <= 0) 
+	    return;
+    extern struct stats stats;
+    static int last_tracked_fd = -1;
```
steadytao: File-descriptor numbers are reusable so a later file can receive the same descriptor and inherit `last_touched_blk` from the previous file. The counter then under-reports that files initial ranges. More generally, one “last block” cannot count unique blocks when writes revisit or move backwards. This needs per-file lifetime state and a representation capable of tracking non-monotonic ranges.

steadytao (changes requested):
Looks quite good to me overall, thank you. Note that CI failures are not because of your changes. Ubuntu is fixed on master, Cygwin will soon be the same so you can rebase then.

fileio.c
seks99x: Yeah totally missed that. I added a reset tracker function and hooked it in receive_data to ensure after every file the fd is reset. Added also an edge case for multiple files and another one for batch mode
Thanks for your review!

main.c
```diff
@@ -1107,6 +1110,8 @@ static int do_recv(int f_in, int f_out, char *local_name)
 
 		write_int(f_out, NDX_DONE);
 		send_msg(MSG_STATS, (char*)&stats.total_read, sizeof stats.total_read, 0);
+		if(protocol_version >= 33)
+		    send_msg(MSG_BLOCK_STATS, (char*)&stats.touched_blocks_4k, sizeof(int64), 0);
```
steadytao: This sends the host-native int64, while io.c decodes the payload with IVAL64(). That produces a byte-swapped value on big-endian systems. Please encode an eight-byte buffer with SIVAL64() before sending it.

main.c
```diff
@@ -443,6 +443,9 @@ static void output_summary(void)
 			human_num(stats.total_transferred_size));
 		rprintf(FINFO,"Literal data: %s bytes\n",
 			human_num(stats.literal_data));
+		if (protocol_version >= 33) 
+             rprintf(FINFO,"Number of modified 4KiB blocks: %s\n",
```
steadytao: "Number of 4 KiB logical blocks touched"

rsync.1.md
```diff
@@ -3547,6 +3537,11 @@ sign) if you want the local shell to expand it.
       Note that this line is only output if deletions are in effect, and only
       if the negotiated protocol is at least 31 (the default when both sides
       are 3.1.0, September 2013, or newer).
+    - `Number of modified 4KiB blocks` is the total number of 4,096-byte boundaries 
+      modified on the receiving disk. This metric decouples network payload from
```
steadytao: This reintroduces the physical-storage claim that #1066 was specifically revised to avoid. Please describe it as the number of unique 4 KiB logical file regions covered by rsyncs own write operations, and avoid “receiving disk” or “disk footprint”.

steadytao (changes requested):
Will fix Ubuntu.

rsync.1.md
seks99x: Yeah idk why I kept treating it as a literal Disk I/O. Fixed

steadytao (changes requested):
`git diff --check` reports trailing whitespace --

The C indentation is inconsistent, the sparse test still describes “physical” blocks/disk writes and the test file lacks a final newline. Please use the logical-region terminology consistently throughout the test as well.

steadytao (approved):

io.c
```diff
@@ -1721,7 +1721,7 @@ static void read_a_msg(void)
 	case MSG_BLOCK_STATS: {
 		char b[8];
 
-		if (msg_bytes != 8 || protocol_version < 33)
+		if (msg_bytes != 8 || protocol_version < 33 || !am_generator)
```
seks99x: it needs to accept generator here to be able to relay it back

birdie-github:
Thanks a ton!

PR #1071, opened 2026-09-02
main: allow --contimeout for daemon connections over --rsh

Zeno-sole:
The --contimeout guard rejected the option whenever connect_timeout was set and the connection was not being made by a socket, so a daemon reached through a remote shell (daemon_connection == 1) was refused the same way as a plain non-daemon remote-shell transfer.  That is how rsync-ssl talks to a daemon: it runs rsync with --rsh pointing at its helper, so "rsync-ssl --contimeout=60 host::mod" died with "The --contimeout option may only be used when connecting to an rsync daemon" before it could even connect.

Only reject --contimeout when there is no daemon connection at all, so the option is accepted for a daemon-via-rsh connection (and still works unchanged for the socket daemon path, which returns earlier through start_socket_client).

Add a regression test that drives a daemon-via-rsh invocation and asserts the option is not rejected, while a plain remote-shell destination is still refused.


steadytao (approved):
CI failures are on master. Ubuntu fixed. Cygwin pending. LGTM.

PR #1075, opened 2026-09-02
daemon: distinguish inetd from local stdin sockets

### LINKED ISSUE #1068: Rsync daemon fails to start if PTY not allocated on Android
I’m running a static build of Rsync for ARM that I built using Nix (`nixpkgs#pkgsCross.aarch64-android.pkgsStatic.rsync`) on an Android phone (specifically a Google Pixel 10 Pro XL running GrapheneOS) via ADB.

Rsync fails to start if it’s launched without a PTY. ADB shell creates a PTY if it’s run without a specific command and thus launches an interactive shell, but doesn’t otherwise (unless you pass -tt). The way I go about driving an ADB shell is usually by launching the noninteractive shell and piping a script to it due to ADB’s strange handling of arguments.

I’m running Rsync with the following command line:
```
/data/local/tmp/rsync --daemon --no-detach --config=/data/local/tmp/rsyncd.conf --verbose --log-file=/data/local/tmp/rsyncd.log
```
The config file reads:
```
address = 127.0.0.1
port = 60873
[user]
path = /storage/emulated/0/
read only = false
```
The log file shows the following error message:
```
2026/08/30 17:31:02 [9693] malformed address : Name does not resolve
2026/08/30 17:31:02 [9693] connect from UNKNOWN ()
2026/08/30 17:31:02 [9693] rsync: [Receiver] failed to set socket option SO_KEEPALIVE: Permission denied (13)
```
Rsync itself prints to stderr:
```
@RSYNCD: 32.0 sha512 sha256 sha1 md5 md4
@ERROR: protocol startup error
```

I’m not sure if this an Rsync bug or an ADB or Android bug.

steadytao:
May fix #1068

## Summary

Require stdin to be a connected IPv4 or IPv6 stream socket before selecting inetd mode.

ADB shell without a PTY presents stdin as a local socket. The existing `is_a_socket(STDIN_FILENO)` check therefore mistakes the ADB process channel for an accepted rsync connection. Rsync then attempts client-address lookup and `SO_KEEPALIVE` on that socket before writing the rsync daemon greeting.

Local sockets are now treated as ordinary process I/O, allowing `--daemon --no-detach` to enter its normal listening loop. Genuine inetd connections remain detected through their connected IP stream and `--no-detach` retains its documented no-op behaviour in inetd mode.

### DIFF
--- a/clientserver.c
+++ b/clientserver.c
@@ -1722,9 +1722,27 @@ static void become_daemon(void)
 	}
 }
 
+/* Inetd supplies a connected IP stream on stdin. Other launchers may use a
+ * local socket for process I/O, which must not select inetd mode. */
+static int is_inet_stream_socket(int fd)
+{
+	struct sockaddr_storage ss;
+	struct sockaddr *sa = (struct sockaddr *)&ss;
+	socklen_t ss_len = sizeof ss;
+	int type;
+	socklen_t type_len = sizeof type;
+
+	if (getpeername(fd, (struct sockaddr *)&ss, &ss_len) < 0
+	 || getsockopt(fd, SOL_SOCKET, SO_TYPE, (char *)&type, &type_len) < 0)
+		return 0;
+
+	return type == SOCK_STREAM
+	    && (sa->sa_family == AF_INET || sa->sa_family == AF_INET6);
+}
+
 int daemon_main(void)
 {
-	if (is_a_socket(STDIN_FILENO)) {
+	if (is_inet_stream_socket(STDIN_FILENO)) {
 		int i;
 
 		/* we are running via inetd - close off stdout and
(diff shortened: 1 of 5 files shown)


schuelermine:
@steadytao I have built this version and tried to run it. Unfortunately, it throws errors about not being able to parse the rsyncd.conf file.
```
Failed to parse config file: /data/local/tmp/rsyncd.conf
```
The file says:
```
address = 127.0.0.1
port = 60873
[user]
path = /storage/emulated/0/
read only = false
```

schuelermine:
Same thing happens if I give it an empty config file.

schuelermine:
(on the Android device, I should say)

PR #1077, opened 2026-09-04
receiver.c: Tighten alt-dest path resolution and partial-dir state validation

seks99x:
This  introduces a couple of architectural best-practices to harden the receiver's state machine and harmonize symlink handling across the delta-basis engine, addressing protocol edge-cases reported by Fyyre (fyyre@fyyre.net).

1. Strict State Validation for partial_dir (receiver.c)
Added explicit validation to ensure `one_inplace` is only triggered when `partial_dir` is actually configured by the client and the `FNAMECMP_PARTIAL_DIR` token is legitimate. This prevents any potential protocol state confusion where a peer sending unexpected/forged tokens could bypass temp-file cleanup or misroute in-place writes.

2. Harmonized Leaf Symlink Handling (receiver.c)
When an operator path was sent, `secure_basis_open()` called `open_no_attacker_symlinks()` for partial-dir/link-dest/fuzzy-dest path resolving, which followed leaf symlinks. Since standard operator-path behavior dictates that leaves should not be followed (consistent with standard do_*_at style, copy_file()) I've updated this logic to enforce O_NOFOLLOW on the final component and walk the parent with `owner_walk_parent`, bringing our path resolution into perfect alignment with standard operator-path behavior.

### DIFF
--- a/receiver.c
+++ b/receiver.c
@@ -119,24 +119,35 @@ static int secure_basis_open(const char *basedir, const char *relpath, int flags
 		return do_open(relpath, flags, mode);
 	}
 
-	/* A peer-supplied --partial-dir basis/staging path (operator_path_resolve set
-	 * by recv_files) may be absolute (module_dir-prefixed on a non-chroot daemon)
-	 * and traverse a symlink the secure_relative_open path can't confine: resolve
-	 * it with the ownership walk, which follows a uid0/euid-owned symlink but
-	 * refuses a foreign one AND (via abspath_excluded_by_module) refuses a target
-	 * the module's exclude hides -- closing the partial-dir exclude bypass. */
-	if (operator_path_resolve) {
-		char fullpath[MAXPATHLEN];
-		const char *p = relpath;
-		if (basedir) {
-			if (pathjoin(fullpath, sizeof fullpath, basedir, relpath) >= sizeof fullpath) {
-				errno = ENAMETOOLONG;
-				return -1;
-			}
+    /* A path with operator_path_resolve set (e.g. an absolute --partial-dir or alt-dest
+     * basis) may traverse a symlink the secure_relative_open path can't confine. We
+     * resolve the operator path with the ownership walk, which safely follows
+     * subdirectory and parent components if they are trusted (uid0/euid-owned) symlinks,
+     * but refuses a foreign one AND (via abspath_excluded_by_module) refuses a target
+     * the module's exclude hides -- closing the partial-dir exclude bypass.
+     * However, we strictly refuse to follow a leaf symlink coming from an operator
+     * path by enforcing O_NOFOLLOW, aligning it on operator-path behavior. */
+    if (operator_path_resolve) {
+        char fullpath[MAXPATHLEN];
+		const char* p = relpath;
+        int dfd, fd, e;
+        const char *leaf;
+        if (basedir) {
+            if (pathjoin(fullpath, sizeof fullpath, basedir, relpath) >= sizeof fullpath) {
+                errno = ENAMETOOLONG;
+                return -1;
+            }
 			p = fullpath;
-		}
-		return open_no_attacker_symlinks(p, flags, mode);
-	}
+        }
+        dfd = owner_walk_parent(p, &leaf);
+        if (dfd < 0)
+            return -1;
+        fd = openat(dfd, leaf, flags | O_NOFOLLOW, mode);
+        e = errno;
+        close(dfd);
+        errno = e; 
+        return fd;
+    }
 
 	/* The confined resolver is needed for the sanitizing daemon
 	 * (am_daemon && !am_chrooted) and for a /./ inner-module chroot
@@ -1085,7 +1096,7 @@ int recv_files(int f_in, int f_out, char *local_name)
 				 * stronger confinement branch in secure_basis_open(), so only
 				 * route the alt-dest basedir read through the walk off-daemon. */
 				if ((basedir && !am_daemon) || fnamecmp_type == FNAMECMP_PARTIAL_DIR)
-					operator_path_resolve = 1;
+				    operator_path_resolve = 1;
 				fd1 = secure_basis_open(basedir, fnamecmp, O_RDONLY, 0);
 				operator_path_resolve = 0;
 			}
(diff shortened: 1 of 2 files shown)


receiver.c
```diff
@@ -119,24 +119,35 @@ static int secure_basis_open(const char *basedir, const char *relpath, int flags
...
+            }
 			p = fullpath;
-		}
-		return open_no_attacker_symlinks(p, flags, mode);
-	}
+        }
+        dfd = owner_walk_parent(p, &leaf);
+        if (dfd < 0)
+            return -1;
+        fd = openat(dfd, leaf, flags | O_NOFOLLOW, mode);
```
steadytao, line 145: This bypasses the existing operator-path symlink policy and unconditionally rejects the leaf. In particular, `--insecure-links` can no longer restore legacy following for an operator-owned leaf symlink while `open_no_attacker_symlinks()` deliberately follows trusted leaves or honours the explicit opt-out. Please preserve that contract and add regressions for a trusted leaf, an untrusted leaf and `--insecure-links`.

steadytao (changes requested):
The `partial_dir` state guard looks reasonable, but the leaf-opening rewrite changes established operator-path and `--insecure-links` behaviour without tests. Please preserve the explicit compatibility opt-out and add focused positive and negative cases before this merges. The diff also needs normal rsync indentation and comment cleanup.

receiver.c
seks99x: I spent a lot of time yesterday tracing these exact code paths, and I'd like to clarify a few mechanics regarding how the leaf is handled here. `--insecure-links` is already preserved,  the legacy following behavior for `--insecure-links` is not broken by this patch. There is an early return check for `symlink_optout_allowed()` at the top of the function that falls back to the raw `do_open()` syscall, which bypasses this O_NOFOLLOW block entirely. `open_no_attacker_symlinks()` function appears to be designed for non-operator paths (like --exclude-from, --read-batch, or --files-from), which typically target specific files. Operator paths behave differently, and the existing comments indicate they are not intended to follow leaf symlinks ( check copy_file() also ).
What "leaf" actually means here for operator paths is that the path is usually concatenated (e.g., basedir + fname). When owner_walk_parent separates the leaf, it is rejecting the destination file itself (e.g., file.txt in /home/Omar/leaf/file.txt), not the parent directory. If the parent directory (leaf/) is a symlink, it is still followed normally if trusted. I've written regression tests that confirm this directory-level leaf symlink traversal works exactly as expected.
Using open_no_attacker_symlinks() here for the leaf file would actually introduce inconsistency. Even if it allowed us to successfully open a symlinked file.txt, subsequent operations (like do_rename_at) would still fail on it, because those downstream syscalls strictly refuse to follow leaf symlinks anyway. Also standard transfers already refuse to follow operator-path fname leaf symlinks, a legitimately 'trusted' symlink is one explicitly configured by the operator as part of the base path itself, not a leaf link planted during the transfer.

steadytao: You are right about `--insecure-links`; I missed the early `symlink_optout_allowed()` return, so that part of my review is resolved. The new tests will still need stronger oracles. Both calls to `run_malicious_sender()` ignore the process result which means an unrelated early protocol failure can pass the test merely because the destination was not changed.

testsuite/strict-basis_test.py
```diff
@@ -0,0 +1,213 @@
...
+
+        count = read_mux_int(peer, buf)
+        for _ in range(3): read_mux_int(peer, buf)
+        for _ in range(count):
+            read_mux_int(peer, buf)
+
+        if iflags & rp.ITEM_TRANSFER:
+            return ndx
+
+def run_malicious_sender(client_args, uri_path, dest_path, test_logic):
```
steadytao: The process result is ignored so an unrelated early protocol failure passes as long as `result_file` is absent. Assert the expected result or diagnostic and add a positive regular-basis control.

testsuite/strict-basis_test.py
```diff
@@ -0,0 +1,213 @@
...
+    body += rp.w_byte(FNAMECMP_BASIS_DIR_LOW) # Force operator-path basis
+    body += rp.w_sum_head(1, len(TARGET_DATA), 16, len(TARGET_DATA)) + rp.w_int(-1) + rp.w_int(0)
+    body += WRONG_MD5
+    out = bytearray(peer.w_ndx(ndx))
+    out += body
+    out += peer.w_ndx(rp.NDX_DONE) * 4
+    peer.send_data(bytes(out))
+
+# We pass --partial so IF it incorrectly reads the symlink, it keeps the file for us to inspect
+run_malicious_sender([f"--link-dest={linkdest}", "--partial"], "mod/", dest, test1_payload)
```
steadytao: The unchanged destination does not prove that the forged partial-dir selector reached and was rejected by the new validation. Assert the expected process outcome and include a control that demonstrates the test fails without the guard.

steadytao (changes requested):
Mostly LGTM. Just some slight test nits.

steadytao (changes requested):
`git diff --check` reports ten whitespace errors and the receiver.c addition does not follow the surrounding C style. Test 2 also still lacks an ordinary non-forged control and does not assert the receiver process result so an unrelated verification failure can satisfy the negative oracle.

steadytao (changes requested):
The current head still fails `git diff --check origin/master...HEAD`, the forged-token test still passes on any unrelated early failure and the leaf-symlink behaviour change remains unrelated and unjustified. Please split that change and address the existing reviews.

seks99x:
Yes thanks I'll fix this now. Regarding the splitting , I felt that too but just didn't want to change the way the reporter wrote the test but that wont work.

steadytao:
No worries. Thank you.

steadytao:
@seks99x just a light bump. If you'd like, I can always assist.

seks99x:
Schedule shifted and i actually travelled today🥲
Actually the test is split and just needs a quick check. The issue is that i won't have device access for another 8–10 hours, so feel free to wrap the tests up so we can save more time for PR #1088 before the release. Thanks!

receiver.c
```diff
@@ -119,35 +119,24 @@ static int secure_basis_open(const char *basedir, const char *relpath, int flags
...
-        if (dfd < 0)
-            return -1;
-        fd = openat(dfd, leaf, flags | O_NOFOLLOW, mode);
-        e = errno;
-        close(dfd);
-        errno = e; 
-        return fd;
-    }
+		}
+		return open_no_attacker_symlinks(p, flags, mode);
```
seks99x: Reverting back to open_no_attacker_symlinks() instead of owner-walk style will keep the alt-dest leaf symlink following. Operator-paths leaf symlinks should not be followed in the destination side.

steadytao (approved):

steadytao:
Any thoughts @seks99x 🤔

seks99x:
Looks good. Great job thanks for that!

PR #1082, opened 2026-09-08
docs: list --password-file among the confined operator paths

samueloph:
The "Symbolic links" section enumerates the operator-supplied paths that are resolved with the ownership walk, but omits --password-file even though getpassf() opens it with open_no_attacker_symlinks() and its own comment states the rule:

    /* --password-file=PATH client open.  Its first line is sent as the
     * auth response, so a planted symlink leaks the target's content
     * (e.g. shadow hashes) to a malicious daemon; the do_stat() * other-access check runs on the target mode and passes 0640 * root:shadow.  Refuse symlinks not owned by uid 0 or our euid. */

So the behaviour is documented in the code but not in the man page, and it is the entry on that list where a missed check has the worst consequence.  A reader checking whether their --password-file is subject to the rule currently finds nothing.

The daemon's "secrets file" is opened through the same helper, but it is an rsyncd.conf parameter rather than a path supplied on the command line, so it is out of scope for this paragraph.


steadytao (approved):
LGTM.

PR #1085, opened 2026-09-09
Fix handling of /dev/std{in,out,err} pseudo-paths and allowing pseudo-paths for overflow uid namespaces

### LINKED ISSUE #1083: [3.5.0 regression] /dev/fd/N from process substitution is refused as an untrusted symlink inside a user namespace
master at 7fef1c82 (includes #1054), Linux.

Inside a user namespace that does not map the host's root uid, /proc/self and /dev/fd stat as the overflow uid, and ona_open() refuses to follow them. The /dev/fd/N path that shell process substitution produces therefore still fails there, although #1054 made it work outside:

```
$ unshare --user --map-root-user --mount --pid --fork --mount-proc bash -c '
    stat -c "%u %n" /proc/self /dev/fd
    rsync -a --exclude-from=<(printf "skip\n") src/ dst/; echo rc=$?'
65534 /proc/self
65534 /dev/fd
refusing to follow a symlink owned by an untrusted user; use --insecure-links locally or "insecure links = yes" in a daemon module ONLY if every path component is trusted
rsync: [client] failed to open exclude file /dev/fd/63: Too many levels of symbolic links (40)
rsync error: error in file IO (code 11) at exclude.c(1719) [client=3.5.0]
rc=11
```

The same command outside the namespace exits 0.

Also reproduced under bubblewrap and in a rootless podman container. Under bubblewrap the sandbox's own /dev/fd is owned by the sandbox user and only /proc/self reports the overflow uid, and the command fails the same way, so the refusal is at /proc/self:

```
65534 /proc/self
1000 /dev/fd
```

#1048 added an exemption for exactly these two links, but only while `pin_transit` is set, i.e. under --confine-root (`int namespace_pin = pin_transit && ...` in syscall.c). The unconfined walk added by #1054 goes through the same two links without it.

Debian builds packages in sbuild's unshare mode, where the pseudo-paths and read-batch-pipe tests fail with the message above.

--insecure-links makes the command succeed, at the cost of disabling the symlink check for every path in that run.

### LINKED ISSUE #1084: [3.5.0 regression] --log-file=/dev/stdout and --files-from=/dev/stdin fail with ENOENT when the descriptor is a pipe
master at 7fef1c82 (includes #1054), Linux. Works with old_versions/rsync_3.4.1 and rsync_3.2.7.

/dev/stdin, /dev/stdout and /dev/stderr are symlinks to /proc/self/fd/N. With a pipe on the descriptor the link target is `pipe:[N]`. The pseudo-path handling from #1054 does not apply, because fd_pin_tail() only recognises paths spelled /proc/self/fd, /proc/<pid>/fd or /dev/fd, so the target is spliced as a relative path and the open fails with ENOENT:

```
$ rsync -a --log-file=/dev/stdout src/ dst/ | cat
rsync: [client] failed to open log-file /dev/stdout: No such file or directory (2)
Ignoring "log file" setting.
```
(exit status 0, nothing logged)

```
$ printf 'file\n' | rsync -a --files-from=/dev/stdin src/ dst/
rsync: failed to open files-from file /dev/stdin: No such file or directory
rsync error: syntax or usage error (code 1) at main.c(1905) [client=3.5.0]
```

The /dev/fd spelling of the same descriptor works:

```
$ rsync -a --log-file=/dev/fd/1 src/ dst/ | cat
2026/09/07 23:44:27 [943874] building file list
2026/09/07 23:44:27 [943874] sent 86 bytes  received 17 bytes  total size 3
```

With stdout redirected to a regular file, --log-file=/dev/stdout works, since the target is then a real path.

The daemon is affected the same way and reports nothing. With `log file = /dev/stdout` in rsyncd.conf and the daemon's stdout a pipe:

```
$ rsync --daemon --no-detach --config=rsyncd.conf --port=12998 --address=127.0.0.1 > >(cat > out) 2> err < /dev/null
```

a client transfer succeeds, but `out` and `err` both stay empty. With old_versions/rsync_3.4.1 the "rsyncd version 3.4.1 starting" line and the connection log arrive in `out`.

Both client commands exit 0 and log / transfer with old_versions/rsync_3.4.1 and old_versions/rsync_3.2.7.

Inside a user namespace that does not map the host's root uid (unshare, bubblewrap, rootless podman), the same paths are additionally refused at /proc/self as an untrusted symlink, with a pipe or a regular file on the descriptor (same cause as #1083).

seks99x:
Using --log-file or --files-from with /dev/stdin, /dev/stdout, or /dev/stderr attached to a pipe previously failed in two ways:

1. On standard hosts, it failed with ENOENT because the symlink target (pipe:[N]) was treated as a relative file path rather than a secure kernel pseudo-path.
2. Inside user namespaces (e.g., rootless podman, unshare), it aborted with ELOOP ("refusing to follow a symlink owned by an untrusted user"). The symlink ownership reported the kernel overflow UID (65534), which completely blocked the walker from following the pseudo-paths.

These crashes occurred due to three overlapping limitations in the symlink path walker:
- fd_pin_tail() only recognized explicit descriptor paths (/dev/fd/N and /proc/self/fd/N), causing the standard stream aliases to fall through to the strict symlink walker.
- The namespace immunity pass (pin_transit) was tied strictly to confine_root. Overflow UID links were only being followed if the transfer was root-confined, causing the strict UID check to block legitimate pseudo-paths in normal, unconfined namespace environments.
- The whitelist for trusted namespace links (namespace_pin) only explicitly exempted /dev/fd and /proc/self, meaning /dev/stdin would have failed the UID check regardless of confinement.

This patch resolves the issues by:
* Updating fd_pin_tail() to natively recognize /dev/std* aliases and map them directly to their corresponding /0, /1, and /2 tails.
* Decoupling pin_transit from confine_root to walk a fine line.
* Expanding namespace_pin to explicitly trust standard stream aliases during the walk, bypassing the overflow UID check that previously blocked these pseudo-paths inside namespaces.
* Adding new cases to the test suite to validate standard stream pseudo-path handling and ensure namespace overflow UID bypasses work correctly without regression.

Fixes: #1083
Fixes: #1084

### DIFF
--- a/syscall.c
+++ b/syscall.c
@@ -162,29 +162,43 @@ static const char *confinement_root(unsigned int *lenp)
  * is not in an fd-pin namespace. */
 static const char *fd_pin_tail(const char *p)
 {
-	const char *s;
+    const char *s;
 
-	if (strncmp(p, "/dev/fd", 7) == 0) {
-		s = p + 7;
-		return (*s == '\0' || *s == '/') ? s : NULL;
-	}
+    /* Group all /dev/ checks under a single prefix comparison */
+    if (strncmp(p, "/dev/", 5) == 0) {
+        s = p + 5;
+        if (strncmp(s, "fd", 2) == 0) {
+            s += 2;
+            return (*s == '\0' || *s == '/') ? s : NULL;
+        }
+        if (strncmp(s, "std", 3) == 0) {
+            s += 3;
+            if (strncmp(s, "in", 3) == 0)
+                return "/0";
+            if (strncmp(s, "out", 4) == 0)
+                return "/1";
+            if (strncmp(s, "err", 4) == 0)
+                return "/2";
+        }
+        return NULL; /* Instantly reject any other /dev/ path */
+    }
 
-	if (strncmp(p, "/proc/", 6) != 0)
-		return NULL;
-	s = p + 6;
-	if (strncmp(s, "self/", 5) == 0)	/* "/proc/self/..." */
-		s += 4;
-	else {					/* "/proc/<pid>/..." */
-		const char *d = s;
-		while (*s >= '0' && *s <= '9')
-			s++;
-		if (s == d || *s != '/')
-			return NULL;
-	}
-	if (strncmp(s, "/fd", 3) != 0)
-		return NULL;
-	s += 3;
-	return (*s == '\0' || *s == '/') ? s : NULL;
+    if (strncmp(p, "/proc/", 6) != 0)
+        return NULL;
+    s = p + 6;
+    if (strncmp(s, "self/", 5) == 0)    /* "/proc/self/..." */
+        s += 4;
+    else {                  /* "/proc/<pid>/..." */
+        const char *d = s;
+        while (*s >= '0' && *s <= '9')
+            s++;
+        if (s == d || *s != '/')
+            return NULL;
+    }
+    if (strncmp(s, "/fd", 3) != 0)
+        return NULL;
+    s += 3;
+    return (*s == '\0' || *s == '/') ? s : NULL;
 }
 
 /* An EXACT pin entry, such as "/proc/self/fd/7" or "/dev/fd/7", whose target is
@@ -347,6 +361,10 @@ static int ona_open(const char *path, int flags, mode_t mode, char *out_abs, siz
 			return -1;
 	}
 
+	/* Tracker: 1 if we are genuinely walking from the system root,
+     * 0 if we are walking a relative path where abspath_step will fake a '/' */
+	int is_anchored = (abspath[0] != '\0');
+
 	/* An fd pin (rrsync rewrites an option path to /proc/self/fd/N so no
 	 * later symlink can redirect it) is spelled outside the root by
 	 * construction, so the walk has to be allowed through /proc/self/fd to
@@ -373,6 +391,7 @@ static int ona_open(const char *path, int flags, mode_t mode, char *out_abs, siz
 			return -1;
 		dfd_owns = 1;
 		abspath[0] = '\0';			/* now resolving from "/" */
+		is_anchored = 1;
 		char *p = remaining;
 		while (*p == '/') p++;
 		memmove(remaining, p, strlen(p) + 1);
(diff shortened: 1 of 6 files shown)


syscall.c
steadytao, line 365: This still conflates two authorities. Whether the walk is inside a recognised fd namespace is independent of whether it may temporarily bypass confinement checks. Introduce separate `fd_namespace_transit` and `confine_pin_transit` state; otherwise future changes can accidentally turn a compatibility exception into a confinement bypass.

testsuite/pseudo-paths_test.py
```diff
@@ -134,5 +136,175 @@
...
+cmd_prefix = shlex.join([str(local_bin), '-a'])
+s_path = shlex.quote(str(ws_src) + '/')
+d_path = shlex.quote(str(ws_dest) + '/')
+
+try:
+    # -------------------------------------------------------------------------
+    # TEST 4: User Namespace Overflow UID (/dev/fd under Overflow UID)
+    # -------------------------------------------------------------------------
+    inner_script = (
+        f'stat -c "Kernel Overflow UID mapping: %u" /proc/self >&2; '
```
steadytao: Assert that this reports the overflow UID before treating the test as coverage for #1083. Otherwise it can pass without exercising the reported failure path.

steadytao (changes requested):
Right direction; wrong implementation. The fd-namespace and confinement states are conflated and the tests do not prove the overflow-UID or daemon cases.

testsuite/pseudo-paths-daemon_test.py
```diff
@@ -0,0 +1,236 @@
...
+
+    conf_host.write_text(f"""pid file = {ws_base}/host.pid
+log file = /dev/stdout
+[test-from]
+path = {dest_host}
+read only = no
+use chroot = no
+uid = 0
+gid = 0
+""")
```
steadytao, line 84: Hard-coding `uid = 0` and `gid = 0` makes this test fail under the unprivileged sanitizer job with setgid failed. Use the current test UID and GID for the host-daemon case or omit the override when it cannot be applied.

testsuite/pseudo-paths-daemon_test.py
```diff
@@ -0,0 +1,236 @@
...
+        daemon_host.wait(timeout=3)
+    except subprocess.TimeoutExpired:
+        try:
+            os.killpg(daemon_host.pid, signal.SIGKILL)
+        except ProcessLookupError:
+            pass
+        daemon_host.wait()
+
+    out_host_data = out_host.read_text() if out_host.exists() else ""
+    err_host_data = err_host.read_text() if err_host.exists() else ""
```
steadytao, line 121: Register daemon termination immediately after `Popen` and perform it from `finally` -- Please also replace the fixed startup sleeps with bounded readiness polling.

testsuite/pseudo-paths_test.py
```diff
@@ -134,5 +136,185 @@
...
+    "import os\n"
+    "proc_uid = os.lstat('/proc/self').st_uid\n"
+    "if proc_uid in (0, os.geteuid()):\n"
+    "    exit(22)\n"
+)
+probe_uid = subprocess.run(launcher + unshare_argv + [sys.executable, '-c', check_ns])
+if probe_uid.returncode == 22:
+    rmtree(base)
+    test_skipped('/proc/self does not expose an overflow uid in this namespace')
+    raise SystemExit(0)
```
steadytao, line 264: This optional namespace subsection calls `test_skipped()` after the portable cases have already passed, reclassifying the entire test as skipped. That causes the Ubuntu skip-oracle failure. Split namespace coverage into its own test or skip only the subsection without marking `pseudo-paths` itself skipped.

steadytao (changes requested):

testsuite/pseudo-paths_test.py
```diff
@@ -134,5 +136,224 @@
...
+
+    symlink_list_ns = ws_base / 'rsync.list'
+    if symlink_list_ns.is_symlink() or symlink_list_ns.exists():
+        symlink_list_ns.unlink()
+
+    # Create the trusted nested symlink pointing to the kernel pipe
+    symlink_list_ns.symlink_to('/dev/stdin')
+
+    # Change the symlink owner to the unmapped host UID (65534) so it maps to 0 inside.
+    os.lchown(symlink_list_ns, 65534, 65534)
```
samueloph: Test 4B fails with `PermissionError` in every unprivileged run of the suite, including Debian's package build, because `os.lchown(symlink_list_ns, 65534, 65534)` needs root; guarding it with `if os.geteuid() == 0:` fixes that, since the chown only matters on the root path where the launcher drops to 65534 before unshare.

```suggestion
    if os.geteuid() == 0:
        os.lchown(symlink_list_ns, 65534, 65534)
```

Other than this, everything else looks good to me as far as I can judge.

seks99x: Thanks. Fixed

steadytao (approved):
LGTM. Much appreciated.

steadytao:
@samueloph bumping your way. Small note that the 3.5.1 regression-ish release is currently planned for the 21st as well for as long we get no more edge/corner cases.

PR #1088, opened 2026-09-14
syscall: follow trusted sender ancestor links

### LINKED ISSUE #1087: Sender-side ELOOP after upgrading to 3.5.0
@peabee reported this in #1067 after upgrading from 3.4.4 to 3.5.0:
```text
rsync: [sender] opendir "/mnt/home/My_Documents" failed: Too many levels of symbolic links (40)
```

Reverting to 3.4.4 reportedly fixes the transfer. The system is Puppy Linux using Slackware64 Current packages. It is not yet clear whether the failure depends on that setup. The command runs inside an application and has not been provided. Peabee says there are no symlinks inside the directory and most files are root:root with a few owned by spot:spot.

I am opening this separately because #1067 was closed after addressing @jb2170’s receiver-side error. #1074 added a diagnostic for an untrusted symlink rejected by `ona_open()`. It did not change the sender’s directory scan. In 3.5.0, that scan can use `secure_opendir()`, whose resolver rejects absolute symlink targets or relative targets that escape its anchor. That could explain this report. It has not been confirmed; having no symlinks inside the directory does not rule it out.

I will investigate the failing path and compare it against 3.4.4 before proposing a fix. Any change needs to retain the protection against symlinks redirecting a transfer outside its permitted root.

@peabee, could you provide the application name and the output of `namei -l /mnt/home/My_Documents` ? There is no need to post the whole directory tree.

steadytao:
Fixes #1087

## Summary

Explicit sender command-line arguments now establish transfer roots and may pass through ancestor symlinks regardless of ownership. This covers absolute and relative directory sources as well as direct file sources used with `--relative`, while keeping recursive traversal confined beneath the selected source root.

`--files-from` entries follow only trusted-owned ancestor symlinks because the source base is operator-selected but individual list entries are not. File-leaf symlinks remain refused unless an existing symlink-following option is enabled.

Daemon confinement remains unchanged.


syscall.c
```diff
@@ -3246,6 +3246,31 @@ int secure_relative_open(const char *basedir, const char *relpath, int flags, mo
...
+		char fullpath[MAXPATHLEN];
+		const char *bname;
+		int dfd, fd, saved_errno;
+		if (snprintf(fullpath, sizeof fullpath, "/%s", relpath) >= (int)sizeof fullpath) {
+			errno = ENAMETOOLONG;
+			return -1;
+		}
+		if (flags & O_DIRECTORY)
+			return open_no_attacker_symlinks(fullpath, flags, mode);
+		dfd = owner_walk_parent(fullpath, &bname);
```
seks99x: When a user explicitly types a sender path on the command line, we should be accepting any symlink without checking whether it is trusted or not.

```
└─# ls -al /home/omar/lnk            
lrwxrwxrwx 1 omar omar 10 Sep 14 14:49 /home/omar/lnk -> /etc/Omar/

┌──(root㉿192)-[~/rsync-latest-news/rsync]
└─# ./rsync -a /home/omar/lnk/ /home/omar/dest/
                                                 
                                                                                                                                                                            
┌──(root㉿192)-[~/rsync-latest-news/rsync]
└─# rm -rf /home/omar/dest/*                   
zsh: sure you want to delete all 2 files in /home/omar/dest [yn]? y
                                                                                                                                                                            
┌──(root㉿192)-[~/rsync-latest-news/rsync]
└─# ./rsync -aR /home/omar/lnk/ /home/omar/dest/
refusing to follow a symlink owned by an untrusted user; use --insecure-links locally or "insecure links = yes" in a daemon module ONLY if every path component is trusted
rsync: [sender] opendir "/home/omar/lnk" failed: Too many levels of symbolic links (40)
rsync error: some files/attrs were not transferred (see previous errors) (code 23) at main.c(1404) [sender=3.5.0-g918ff435]
                                                                                                                                                                            
┌──(root㉿192)-[~/rsync-latest-news/rsync]
└─# 
```

Relative paths symlinks (with -R/--relative) still not fixed. Full paths containing trusted symlinks with --relative/-R flag is the only case covered on this patch.

```

┌──(root㉿192)-[~/rsync-latest-news/seks99x/batch/rsync]
└─# ./rsync -aR lnk/ /home/omar/dest/
rsync: [sender] opendir "/root/rsync-latest-news/seks99x/batch/rsync/lnk" failed: Too many levels of symbolic links (40)
rsync error: some files/attrs were not transferred (see previous errors) (code 23) at main.c(1404) [sender=3.5.0-g08102405]
                                                                                                                                                                            
┌──(root㉿192)-[~/rsync-latest-news/seks99x/batch/rsync]
└─# ls -al lnk          
lrwxrwxrwx 1 root root 10 Sep 15 01:14 lnk -> /etc/omar/
                                                                                                                                                                            
┌──(root㉿192)-[~/rsync-latest-news/seks99x/batch/rsync]
└─# 
```
Append also cases on the testsuite covering both of these edge cases.

seks99x (changes requested):

syscall.c
steadytao: Cheers.

seks99x:
Great job! there is still another case not covered. File transfers using the `-R` (`--relative`) flag are still broken across the board if a symlink is included in the path (full/relative paths). Directories or recursive transfers works well with full/relative paths.

```jsx

┌──(root㉿192)-[~/rsync-latest-news]
└─# rsync/rsync -R rsync/lnk/test/a /home/omar/dest/ -> rel/full path case
rsync: [sender] send_files failed to open "/root/rsync-latest-news/rsync/lnk/test/a": Too many levels of symbolic links (40)
rsync error: some files/attrs were not transferred (see previous errors) (code 23) at main.c(1404) [sender=3.5.0-gd0a5ab7f]

Expected Behavior (rsync 3.4.1)
seks@omars-Mac-mini rsync-3.4.1 % ./rsync -R lnk/sub/a dest/ 
seks@omars-Mac-mini rsync-3.4.1 % ls dest/
lnk

```

peabee:
> This is built against TrixiePup64, so it may not run on your Slackware-based Puppy - [rsync.tar.gz](https://github.com/user-attachments/files/32215614/rsync.tar.gz)
> 
> Make sure to keep your existing rsync; if this binary reports a library error, send me that error and your exact Puppy version/environment so I can provide a matching build.

Sorry for the delay in responding....
Many thanks for the binary - appreciated - all dependencies were satisfied.
I can confirm that with this version, the backup to /mnt/home/My_Documents where /mnt/home is a symlink works successfully - no errors.
Looking forward to the new version appearing.
Thanks.

steadytao:
Much appreciated. The new version should be released on the 21st for as long we catch no more edge/corner cases.

seks99x:
``files-from`` uses relative by default and still broken. They should be following only trusted symlinks because we are inside the operators-tree. 

```jsx
┌──(root㉿192)-[~/rsync-latest-news/rsync]
└─# ./rsync -av --files-from=/root/rsync-latest-news/rsync/files.txt /root/rsync-latest-news/seks99x/batch/rsync/  /home/omar/dest/
building file list ... done
rsync: [sender] send_files failed to open "/root/rsync-latest-news/seks99x/batch/rsync/lnk/test/a": Too many levels of symbolic links (40)

sent 185 bytes  received 31 bytes  432.00 bytes/sec
total size is 1,057  speedup is 4.89
rsync error: some files/attrs were not transferred (see previous errors) (code 23) at main.c(1404) [sender=3.5.0-gc9f8e21f]
                                                                                                                                                                            
┌──(root㉿192)-[~/rsync-latest-news/rsync]
└─# ls -al lnk                                                                                                                     
lrwxrwxrwx 1 root root 10 Sep 15 18:34 lnk -> /etc/omar/
                                                                                                                                                                            
┌──(root㉿192)-[~/rsync-latest-news/rsync]
└─# cat files.txt            
lnk/test/a. -> contains symlink

Expected Behavior (rsync 3.4.1)
seks@omars-Mac-mini rsync-3.4.1 % ./rsync -av --files-from=files.txt . dest/
building file list ... done
lnk/
lnk/sub/
lnk/sub/a

sent 161 bytes  received 41 bytes  404.00 bytes/sec
total size is 3  speedup is 0.01

seks@omars-Mac-mini rsync-3.4.1 % cat files.txt 
lnk/sub/a -> contains symlink
                                                                                                                                                                            
```

steadytao:
Ahhhh true. Thank you.

flist.c
```diff
@@ -2028,10 +2139,14 @@ static void interpret_stat_error(const char *fname, int is_dir)
  * Returns NULL with errno set on failure, like opendir(). */
 static DIR *secure_opendir(const char *fbuf)
 {
-	int dfd, fl;
+	int dfd, fl, matched;
 	DIR *d;
 
-	if (am_daemon && (!am_chrooted || module_dirlen)
+	if (!am_daemon && am_sender
```
seks99x: `secure_relpath_active()` would be better / more accurate on the non-daemon sender case ?

```
	if (!secure_relpath_active()
	 && (dfd = open_sender_source_path(fbuf, O_RDONLY | O_DIRECTORY, &matched), matched)) {
```

seks99x:
If a directory listed in --files-from contains or leads through a trusted symlink, traversing it fails . Trusted symlinks in the path should be followed correctly here.
Files are working right, if I changed the path to lnk/test/a it will work as expected.

```
┌──(root㉿192)-[~/rsync-latest-news/rsync]
└─# ./rsync -av --files-from=/root/rsync-latest-news/rsync/fi . /home/omar/dest/
building file list ... 
rsync: [sender] opendir "/root/rsync-latest-news/rsync/lnk/test" failed: Too many levels of symbolic links (40)
done
lnk/
lnk/test/

sent 84 bytes  received 22 bytes  212.00 bytes/sec
total size is 0  speedup is 0.00
rsync error: some files/attrs were not transferred (see previous errors) (code 23) at main.c(1404) [sender=3.5.0-ge7a40140]
                                                                                                                                                                            
┌──(root㉿192)-[~/rsync-latest-news/rsync]
└─# cat fi                
lnk/test/
                                                                                                                                                                            
┌──(root㉿192)-[~/rsync-latest-news/rsync]
└─# ll lnk
lrwxrwxrwx 1 root root 10 Sep 15 18:34 lnk -> /etc/omar/
                                                                                                                                                                            
┌──(root㉿192)-[~/rsync-latest-news/rsync]
└─# 
```

seks99x (changes requested):

steadytao:
Oof ye. Okay give me more time on this. Good to get a rough shape and it semi-working but it needs a lot more for proper coverage.

seks99x:
Take your time, mate! Symlinks are an absolute mess in this codebase😂

steadytao:
Symlinks are an absolute mess everywhere. I need to start a revolt!!!

flist.c
```diff
@@ -2160,13 +2213,20 @@ static void interpret_stat_error(const char *fname, int is_dir)
...
 static DIR *secure_opendir(const char *fbuf)
 {
 	int dfd, fl, matched;
 	DIR *d;
 
-	if (!am_daemon && am_sender
+	if (filesfrom_owner_walk_active()) {
+		/* The source base is operator-selected, while each list entry may not
+		 * be. Follow only trusted-owned symlinks while opening the directory. */
+		dfd = open_no_attacker_symlinks(fbuf, O_RDONLY | O_DIRECTORY, 0);
```
seks99x, line 2265: Append operator-path-resolve here. if root is confined we shouldn't be even following the link if trusted and out of the tree.
```
	if (filesfrom_owner_walk_active()) {
		/* The source base is operator-selected, while each list entry may not
		 * be. Follow only trusted-owned symlinks while opening the directory. */
		int saved = operator_path_resolve; 
        operator_path_resolve = 1;
		dfd = open_no_attacker_symlinks(fbuf, O_RDONLY | O_DIRECTORY, 0);
		operator_path_resolve = saved;
```
Behavior:
```jsx
with operator-path-resolve

└─# ./rsync -av --files-from=/home/omar/files.txt /home/omar/ /home/omar/dest/
building file list ... done
created directory /home/omar/dest
lnk/
lnk/.gitattributes
lnk/.gitignore

┌──(root㉿192)-[~/rsync-latest-news/rsync]
└─# ./rsync -av --confine-root=/home/omar/ --files-from=/home/omar/files.txt /home/omar/ /home/omar/dest/
building file list ... 
rsync: [sender] opendir "/home/omar/lnk" failed: Too many levels of symbolic links (40)
done
lnk/

sent 56 bytes  received 19 bytes  150.00 bytes/sec
total size is 0  speedup is 0.00
rsync error: some files/attrs were not transferred (see previous errors) (code 23) at main.c(1404) [sender=3.5.0-g4f6d572e]
                                                                                                                            
without operator-path-resolve  

┌──(root㉿192)-[~/rsync-latest-news/rsync]
└─# ./rsync -av --confine-root=/home/omar/ --files-from=/home/omar/files.txt /home/omar/ /home/omar/dest/
building file list ... done
lnk/.gitattributes
lnk/.gitignore
                                                                                                                                                                                                                   
```

sender.c
```diff
@@ -710,16 +682,12 @@ void send_files(int f_in, int f_out)
...
-			 * governs the leaf so a raced leaf symlink is refused.  A
-			 * symlink-following mode (-L/--copy-unsafe-links/-k) or
-			 * --insecure-links keeps the legacy open below. */
+			/* A files-from entry follows only trusted-owned ancestors because
+			 * its source base is operator-selected but the entry itself may not
+			 * be. Other paths stay confined beneath their explicit transfer root.
+			 * Every file leaf remains O_NOFOLLOW. */
 			if (files_from) {
-				fd = sender_open_filesfrom(fname, O_RDONLY);
+				fd = do_open_checklinks(fname);
```
seks99x, line 693: `do_open_checklinks` uses `do_open_nofollow` which only refuse to follow leaf symlinks. By appending this if a source path uses a files-from option, a sub directory could be raced and swapped by a symlink which will let an attacker redirect the transfer out of the tree.
pinning the `fname` parent and refuse to follow the leaf if no *-links option is occurred would be better here.

flist.c
```diff
@@ -327,7 +327,7 @@ int open_sender_source_path(const char *path, int flags, int *matched)
 		return -1;
 
 	*matched = 1;
-	rootfd = open(best->path, O_RDONLY | O_DIRECTORY);
+	rootfd = open_anchor_dirfd(best->path);
```
seks99x: Could we cache the validated rootfd ?

`open_sender_source_path()` currently opens and fstats `best->path` for every file. The source root (best->path) remains unchanged for the lifetime of the transfer and also it is the trusted path coming from the user, so the validated `rootfd` is already the trusted anchor. the untrusted `rel` only is the one be better resolved on every file entry.

seks99x (changes requested):

steadytao:
Bit of a scope explode but I do agree it is quite necessary. Files-from directory walks now enforce `--confine-root`, including trusted links outside or above the root. Content opens use the pinned parent with an `O_NOFOLLOW` leaf and retain `O_NOATIME`. `--remove-source-files` uses the same parent walk. Source-root validation now keeps one validated root fd at a time rather than reopening it for every file.

Added regression coverage for trusted in-root links, out-of-root and ancestor links, untrusted directory links and files-from removal. The focused Cygwin suite passes.

seks99x:
Yeah thats a lot!
Great job mate. Regarding the `--relative` and `files-from`, testing & security phases LGTM. I'll still need to take some more time reviewing the new changes on `syscall.c`.
I hope I get the time to finish this tonight, this is necessary to be merged before the new release.

steadytao:
Of course. Take your time.

seks99x:
LGTM. Thank you for this hard work.

steadytao:
Much appreciated.

peabee:
Tested the Slackware-Current build of rsync-3.5.1 and can confirm all OK with my backup scenario. Thanks.

steadytao:
Thank you very much @peabee !

